PR #1, opened 2015-03-26
Wqp 626

mbucknell:
Initial WQP UI flask application set up. Right now this creates an empty home page and an empty contact us page. All common page elements have been put into a base.html page which all other pages will extend.

Moved the img and less directories.


mbucknell:
I believe this now includes a complete base.html template as well as functioning index and contact_us as well as all of the static files needed by the application. Since this is a nice working chunk. I'm going to go ahead and merge.

PR #33, opened 2015-04-17
one url per page

jkreft-usgs:
We don't want multiple versions of the same page, so killing off the .jsp in favor of cool urls


mbucknell:
:+1:

PR #44, opened 2015-05-27
serve kml style snippet and kml icon images

jkreft-usgs:


portal_ui/templates/wqp_styles.kml
```diff
@@ -0,0 +1,65 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<kml xmlns="">
+<Document id="wqp_styles">
+	<Style id="normalARSSite">
+		<IconStyle>
+			<Icon>
+				<href>../img/orangecir.png</href>
```
mbucknell: Does this work with relative urls? Seems like they should be absolute. I know the old one uses relative, but at least I don't get any symbols on google map and google earth only shows yellow pins.

jkreft-usgs: this is just a snippet that is then manipulated further by the
kml-producing java code.  The absolute URLs are inserted at that time.

Jim Kreft
Center for Integrated Data Analytics
U.S. Geological Survey
8505 Research Way
Middleton, WI  53562
608-821-3919
jkreft@usgs.gov

mbucknell:
:+1:

PR #48, opened 2015-06-15
canonical links, form autocomplete to false

jkreft-usgs:
This will hopefully clue chrome (and other browsers) into the fact that we don't actually want autocomplete


mbucknell:
:+1:

PR #60, opened 2016-01-07
Updated the python dependencies to the latest version. Made one small…

mbucknell:
… change to prevent a warning from being displayed in the log.


mbucknell:
Not much to review here so I'm going to go ahead and merge.

PR #61, opened 2016-01-14
Wqp 744

mbucknell:
Updated javascript libraries used in the portal page with the exception of OpenLayers.

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,0 +1,3 @@
+{
+  "directory" : "portal_ui/bower_components"
+}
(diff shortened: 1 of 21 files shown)


mbucknell:
@mamcderm-usgs, @ayan-usgs , @instantsoup , and @dsteinich : Take a look, if you want take a look. Unfortunately I have updated some files to use tab spacing so it makes a code review challenging. The majority of the changes are in utils.js, placeSelects.j, portalModels.js, and portalViews.js and their tests.

I need to update the Jenkins jobs to bring in the javascript dependencies. Once I do that, I will merge. Before I do anything else on this project, I will reformat all javascript files to use tabs so code reviews can actually happen.

PR #67, opened 2016-01-19
Refactor

mbucknell:
Primarily refactored the way the Place selects are managed. The file placeInputView.js replaces placeSelects. Code functionally does the same thing but the model objects are no longer global objects. They are created in onReady.js and passed into placeInputView as is the container for the place inputs. The code in the file itself was streamlined and only the initialize functions is made public.

I also added a testConfig.js file that is loaded when the tests are loaded and provides initialization for the Config object. I will need to clean up the tests at some point to not assign Config properties in the tests.


mbucknell:
Added separate view code to set up the Point Location and BoundingBox sections of the input form.

PR #76, opened 2016-02-17
Wqp 746

mbucknell:


mbucknell:
OpenLayers 2 is still in the project because it's still used for the coverage mapper page. I moved ol3 specific map utilities into the ol3Utils subdirectory.

portal_ui/static/js/MapConfig.js
```diff
@@ -0,0 +1,45 @@
+/* jslint browser: true */
```
mbucknell: Moved defaults into a separate MapConfig.js that both the ol2 and ol3 code can access.

portal_ui/static/js/identifyDialog.js
```diff
@@ -0,0 +1,139 @@
+/* jslint browser: true */
```
mbucknell: Now using Handlebars to create the html for the site info and hidden form inputs. This module handles both the UI dialog and the popover. The popover is used when the width of the screen is extra small.

portal_ui/static/js/identifyDialog.js
```diff
@@ -0,0 +1,139 @@
...
+
+	var FEATURE_INFO_TEMPLATE = Handlebars.compile(
+		'{{#if exceedsFeatureLimit}}' +
+		'Retrieved {{features.length}} sites, only showing ' + FEATURE_LIMIT + '<br />' +
+		'{{/if}}' +
+		'{{#each features}}' +
+			'<br /><table>' +
+			'<tr><th>Station ID: </th><td class="details-site-id">{{values_.name}}</td></tr>' +
+			'<tr><th>Name: </th><td>{{values_.locName}}</td></tr>' +
+			'<tr><th>Type: </th><td>{{values_.searchType}}</td></tr>' +
```
dsteinich: Is this the correct site type to show - it matches what was selected on the UI form. type is more descriptive.

mbucknell: I'll double check with Jim.

PR #82, opened 2016-02-24
Fixed bug in how infinite scrolling was implemented. Added a test to …

mbucknell:
…test for the correct result if more data needs to be fetched.


mbucknell:
Nothing really to review ... I just needed to follow the example in the select2 documentation.

PR #85, opened 2016-03-17
Fixed an issue where on small devices, the bottom part of the downloa…

mbucknell:
…d form was not showing. Also fixed a couple of other layout issues.

I am now using min-width rather than min-device-width and using the bootstrap variables for the various size breakpoints.

### DIFF
--- a/portal_ui/static/less/portal_form.less
+++ b/portal_ui/static/less/portal_form.less
@@ -1,17 +1,17 @@
 // Styling specific to the portal form
-#params .panel {
+#params .panel:not(#map-and-query-div)  {
   .panel-heading {
     button {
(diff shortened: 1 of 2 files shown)


portal_ui/static/less/portal_form.less
```diff
@@ -1,17 +1,17 @@
 // Styling specific to the portal form
-#params .panel {
+#params .panel:not(#map-and-query-div)  {
```
mbucknell: This was the fix for the issue in WQP-790. The map-and-query-div is now part of the #params form but we don't want to hide it on small screens. So added the :not selector to prevent the style from being applied to that div.

PR #86, opened 2016-03-21
Wqp 793

mbucknell:
Added the ability to generate a legend from a GetLegendGraphic call. The legend is now on the map, unless on an extra small device. Then it is below the map.

### DIFF
--- a/portal_ui/static/js/views/siteMapView.js
+++ b/portal_ui/static/js/views/siteMapView.js
@@ -40,6 +40,7 @@ PORTAL.VIEWS.siteMapView = function(options) {
 		});
 
 		var $mapContainer = options.$container.find('#query-map-container');
+		var $legendContainer = options.$container.find('#query-map-legend-div');
 		var $map = options.$container.find('#' + mapId);
 		var $showHideBtn = options.$container.find('.show-hide-toggle');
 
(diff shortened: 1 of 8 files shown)


portal_ui/static/less/portal_form.less
```diff
@@ -226,111 +226,116 @@
 }
 
 // Styles for map and query section
-#map-ops-button-div {
-  .make-row();
-  > div {
-    padding-bottom: 10px;
-  }
-  #show-on-map-button-div {
```
mbucknell: In addition to adding the styling for the new legend and removing the styling for the old, reorganized the styles associated with the mapping-div so that they are easier to find.

portal_ui/static/js/views/siteMapView.js
```diff
@@ -21,22 +21,26 @@ PORTAL.VIEWS.siteMapView = function(options) {
 
 	var mapId = 'query-results-map';
 
-	var identifyDialog = PORTAL.VIEWS.identifyDialog({
```
mbucknell: Moved the calls to create the identifyDialog and portalDataMap objects to initialize. These objects shouldn't be initialized until the user calls initialize for this view.

mbucknell:
I'm going to merge. I have another very simple change ready to go and I don't want to have to merge changes to portal.html.

PR #91, opened 2016-04-12
Wqp 803

mbucknell:
Added two leaflet maps. The "inset" map is shown within the site parameters box and contains zoom control, a button to "expand" the map, and a picker to pick the nav type. If a nav type is picked or the expand button is clicked, the inset map is hidden and the full size map is shown. In addition to the above, the expand button is now a collapse button and the map also contains a layer switcher. If the nav type is changed to the null value, the map is hidden and the inset map shown again.

I did not write any tests for nldiMapViews. It seemed tot brittle and artificial to set up tests for it.


bower.json
```diff
@@ -17,17 +17,21 @@
...
-    "bootstrap": "~3.3.6",
-    "jquery": "~2.2.0",
-    "select2": "~4.0.1",
-    "jquery-ui": "~1.11.4",
-    "numeral": "~1.5.3",
+    "bootstrap": "^3.3.6",
+    "jquery": "^2.2.3",
+    "select2": "^4.0.2",
+    "jquery-ui": "^1.11.4",
+    "numeral": "^1.5.3",
```
dsteinich: per our conversation - double check which is more appropriate, ~ or ^

portal_ui/static/js/leafletControls/NldiNavControl.js
```diff
@@ -0,0 +1,49 @@
...
+	onAdd : function(map) {
+		"use strict";
+
+		var container = L.DomUtil.create('div', 'leaflet-nldi-nav-picker-div');
+		this._selectEl = L.DomUtil.create('select', 'leaflet-nldi-nav-picker', container);
+		this._selectEl.innerHTML = '<option></option>' +
+				'<option value="UM">Upstream main</option>' +
+				'<option value="DM">Downstream main</option>' +
+				'<option value="DD">Downstream with diversions</option>' +
+				'<option value="DD">Downstream with diversions</option>';
```
dsteinich: UT - Upstream tributary?

bower.json
mbucknell: I think tilde is more appropriate as that only allows patch updates. I'm not sure why it changed to using ^. I'll do a little more investigating. The safest thing to do would be to lock down the dependencies and not use a ^ or ~. What are  your thoughts?

portal_ui/static/js/leafletControls/NldiNavControl.js
mbucknell: Thanks .. will fix.

bower.json
dsteinich: If lock down = consistent/repeatable, that sounds like a good thing. Especially if not lock down = you and I might see different things (or possible differences between development & release).

PR #97, opened 2016-04-19
Wqp 806

mbucknell:
Using the count endpoint and a post call to retrieve query counts now.

For now, the query form still uses the get method. When the post method is available we should only have to update portal.html to use post for the method on the form.


mbucknell:
This has all the changes except for making the query form submission a post.

PR #98, opened 2016-04-19
site pages!

jkreft-usgs:
This gives us a restful URI for each site page, and passes that information off to a cache so that we don't DOS ourselves


portal_ui/templates/provider_base.html
```diff
@@ -0,0 +1,17 @@
...
+    <h1>Water Quality Portal Data Providers</h1>
+
+    <p>Learn more about Water Quality Portal Data Providers <a href="{{ url_for('wqp_description-canonical') }}">here</a>.</p>
+
+    {% for provider in providers %}
+        <a href="{{ url_for('uri_provider', provider_id = provider) }}">{{ provider }}</a></br>
+    {% endfor %}
+{% endblock %}
+
+{% block body_page_script %}{% endblock %}
```
mbucknell: You don't need to put this here, but its not harmful.

portal_ui/templates/site.html
```diff
@@ -0,0 +1,68 @@
...
+        {% if site['SourceMapScaleNumeric'] %}The source map scale is 1:{{ site['SourceMapScaleNumeric'] }}.{% endif %}
+        {% else %} No horizontal location accuracy metadata is available.
+    {% endif %}
+        {% elif site['StateName'] and site['CountyName']  %}This site is located in {{ site['CountyName'] |title}} County, {{ site['StateName']|title }}
+    {% endif %}
+    {% if  site['VerticalMeasure/MeasureValue'] and site['VerticalMeasure/MeasureUnitCode']%} This site is at an elevation of {{ site['VerticalMeasure/MeasureValue'] }} {{ site['VerticalMeasure/MeasureUnitCode'] }}
+    {% if site['VerticalAccuracyMeasure/MeasureValue'] and site['VerticalAccuracyMeasure/MeasureUnitCode'] and site['VerticalCollectionMethodName']  %}
+    and the accuracy of the elevation measurement, collected using the method "{{ site['VerticalCollectionMethodName'] }}" is {{ site['VerticalAccuracyMeasure/MeasureValue'] }} {{ site['VerticalAccuracyMeasure/MeasureUnitCode'] }}.
+    {% if site['VerticalCoordinateReferenceSystemDatumName'] %} The vertical coordinate reference system is {{ site['VerticalCoordinateReferenceSystemDatumName'] }}.{% endif %}
```
mbucknell: Consider using tabbing to indicate where {% if %}{% endif %} blocks are ... particularly when you have nesting of {% if} %} blocks.

portal_ui/utils.py
```diff
@@ -49,3 +59,189 @@ def geoserver_proxy_request(target_url):
...
     return make_response(resp.content, resp.status_code, resp.headers.items())    
 
+
+def generate_provider_list(endpoint):
+    """
+
+    :param endpoint: the base codes endpoint
+    :return: a list of provider names
+    """
+    provider_endpoint = endpoint+'/providers'
```
mbucknell: Style issue: should put spaces around the '+' operator. This is what PEP 8 standard says.

portal_ui/utils.py
```diff
@@ -49,3 +59,189 @@ def geoserver_proxy_request(target_url):
...
+    r = requests.get(provider_endpoint, {"mimeType": "json"})
+    status_code = r.status_code
+    provider_list = None
+    if status_code == 200:
+        codes = r.json().get('codes')
+        provider_list = []
+        for code in codes:
```
mbucknell: Consider using a list comprehension.

portal_ui/utils.py
```diff
@@ -4,6 +4,10 @@
 
 from bs4 import BeautifulSoup
 from flask import request, make_response
+import tablib
+import ujson
```
mbucknell: Is there a reason to use ujson rather than the standard python json parser?

portal_ui/utils.py
```diff
@@ -49,3 +59,189 @@ def geoserver_proxy_request(target_url):
...
+def check_org_id(org_id, code_endpoint):
+    org_exists = False
+    org_endpoint = code_endpoint+'/Organization'
+    r = requests.get(org_endpoint, {"mimeType": "json", "text":org_id} )
+    status_code = r.status_code
+    if status_code == 200:
+        codes = r.json().get('codes')
+        for code in codes:
+            if code['value'] == org_id:
+                org_exists = True
```
mbucknell: Consider using break to break out of for loop once org_exists = True

portal_ui/utils.py
```diff
@@ -49,3 +59,189 @@ def geoserver_proxy_request(target_url):
...
+    """
+
+    :param base_url: the base url we are using for the generating the search URL
+    :param organization_id:
+    :param provider_id:
+    :return: a list of dicts that describe sites that are associated with an organization under a data provider
+    """
+    search_endpoint = base_url+"Station/search/"
+    r = requests.get(search_endpoint, {"organization": organization_id, "providers": provider_id,
+                                       "mimeType": "geojson", "sorted": "no", "uripage": "yes"})
```
mbucknell: What does the uripage parameter do? Just curious.

portal_ui/views.py
```diff
@@ -1,9 +1,23 @@
-import requests
-
-from flask import render_template, request, make_response, redirect, url_for
-
+from flask import render_template, request, make_response, redirect, url_for, abort
 from . import app
 from .utils import pull_feed, geoserver_proxy_request
+from utils import generate_provider_list, generate_organization_list, get_site_info, check_org_id, \
```
mbucknell: Should change this back to using relative (put the . in front of utils)/

portal_ui/views.py
```diff
@@ -1,9 +1,23 @@
...
+    make_cache_key, generate_site_list_from_csv, generate_redis_db_number
+from flask.ext.cache import Cache
+import redis
+import ast
+
+
+
+# set some useful local variables from the global config variables
```
mbucknell: Some of these are being passed to the templates as template variables. The config variables are available in the templates as {{ config.CONSTANT_NAME }}. I would prefer that this be used. Then it is clear that these are configuration variables.

jkreft-usgs: I changed robots.txt

portal_ui/utils.py
jkreft-usgs: break added

portal_ui/utils.py
jkreft-usgs: ujson is 5-10x faster than the standard json parser for large datasets (it drops into C)

portal_ui/utils.py
jkreft-usgs: good point

portal_ui/utils.py
jkreft-usgs: done

portal_ui/utils.py
mbucknell: Have to weigh bringing in a library versus whether the speedup is necessary.

portal_ui/views.py
```diff
@@ -154,3 +168,140 @@ def kml():
 def images(image_file):
     return app.send_static_file('img/'+image_file)
 
+
+@app.route('/provider/', endpoint='uri_base')
+def uri_base():
+    providers = generate_provider_list(code_endpoint)
+    if providers['status_code'] == 200 and providers['providers']:
```
mbucknell: As we talked about, we should consider whether the utility should make the service call or whether the service call should be made directly in the view and have the utilitiy process the successful service call. I favor the later, because writing tests for it is simpler and you keep error handling completely within the view.

portal_ui/templates/Provider.html
```diff
@@ -0,0 +1,15 @@
+{% extends "base.html" %}
```
mbucknell: Can we make this 'provider.html' (lowercase)?

mbucknell:
Looks like its ready to go 👍

portal_ui/utils.py
jkreft-usgs: In this case, it really cut the page load times.  However, at this point, I'm not using the import from json functionality.  Most likely, this util (and its library) will go away soon.

portal_ui/templates/site.html
jkreft-usgs: yeah, that was confusing

portal_ui/views.py
jkreft-usgs: that makes sense.  I will work on re-refactoring over the next few weeks

PR #100, opened 2016-04-21
only create redis session once

jkreft-usgs:
Only creating the redis session once, instead of multiple times, massively improves redis interaction speed


mbucknell:
Looks good to go

PR #102, opened 2016-04-22
Wqp 808

mbucknell:
Added a distance input and clear button to NldiControl. This was named NldiNavControl, but since other inputs are in the control, changed to the more generic NldiControl. Updated the tests for the new functionality.

Also started saving the comid in nldiView so that the user can change the navigation direction or distance and have the current comid used for a new query. As a result, also added a clear button to clear the saved comid, inputs and layers on the NLDI map.


mbucknell:
Going to go ahead and merge so that I can get all of the changes in. Feel free to still comment.

jkreft-usgs:
I would just like to note that this works incredibly well

PR #106, opened 2016-04-25
use streaming to populate cache

jkreft-usgs:
streaming instead of using the giant CSV.  This also allows me to use the geojson endpoint instead of the csv


mbucknell:
Looks good 👍

PR #107, opened 2016-04-25
yank flask-cache

jkreft-usgs:
flask-cache doesn't work with a redis password, and we need a password, so...


jkreft-usgs:
see https://github.com/thadeusb/flask-cache/issues/130

PR #111, opened 2016-04-26
put a map on the organization pages

jkreft-usgs:


portal_ui/templates/sites.html
```diff
@@ -2,14 +2,99 @@
...
+                        fillOpacity: 0.7
+                    });
+            }
+            });
+        dataLayer.addTo(map);
+        map.fitBounds(L.geoJson(data).getBounds());
+    }
+
+    var pageURL = "{{ config.LOCAL_BASE_URL }}{{ request.path }}";
+    $.getJSON( pageURL, {
```
mbucknell: I don't like that you have to essentially get the same information that is already in a template variable. I understand why you are doing it, but I don't like it.

portal_ui/views.py
```diff
@@ -218,31 +218,47 @@ def uri_organization(provider_id, organization_id):
...
+    search_endpoint = base_url + "Station/search/"
     if redis_config:
         redis_db_number = generate_redis_db_number(provider_id)
         redis_session = redis.StrictRedis(host=redis_config['host'], port=redis_config['port'],
                                           db=redis_db_number, password=redis_config.get('password'))
         all_sites_key = 'all_sites_' + provider_id + '_' + organization_id
         redis_all_site_data = redis_session.get(all_sites_key)
         if redis_all_site_data:
-            sites_list = pickle.loads(redis_all_site_data)
+            print("getting from cache")
```
mbucknell: You shouldn't leave print statements in. We should probably add logging to this application.

portal_ui/templates/sites.html
jkreft-usgs: yeah, if it didn't basically double the page size sent over the wire, I wouldn't make the second call, but it significantly decreases page load to get the geojson for the map asyncronously.

portal_ui/views.py
jkreft-usgs: removed print statements in later commit

PR #118, opened 2016-05-13
bing webmaster tools

jkreft-usgs:


jkreft-usgs:
I want to tell bing so that it will show up in usa search as well

PR #119, opened 2016-05-19
add an announcement banner for shutdown

jkreft-usgs:
this will allow for quickly changing an announcement.  javascript should mean that (hopefully) there isn't stuff ingested into google for weeks


mbucknell:
👍

PR #120, opened 2016-06-02
upstream-downstream display on the site page

jkreft-usgs:


mbucknell:
Just a general comment. It would be nice if the lines weren't so long. It is very difficult to look at them in git since many exceed the width of the viewing window. Also consistent indentation helps in reading the code. These are both maintenance type issues. I think the work is good to go.

jkreft-usgs:
I agree, and we I expect to clean things up more (and also move the js content into a webassets file!)

PR #122, opened 2016-06-30
Wqp 833

mbucknell:
Rather than putting all of the sites returned from the NLDI service call into the query form, use the new nldiurl parameter. Allows removal of the limit of 50 sites.

### DIFF
--- a/portal_ui/static/js/views/downloadFormView.js
+++ b/portal_ui/static/js/views/downloadFormView.js
@@ -79,7 +79,7 @@ PORTAL.VIEWS.downloadFormView = function(options) {
 		var nldiView = PORTAL.VIEWS.nldiView({
 			insetMapDivId : 'nldi-inset-map',
 			mapDivId : 'nldi-map',
-			$siteInputContainer : options.$form.find('#nldi-sites-container')
+			$inputContainer : options.$form.find('#nldi-param-container')
 		});
 		var samplingParametersInputView = PORTAL.VIEWS.samplingParameterInputView({
 			$container : options.$form.find('#sampling'),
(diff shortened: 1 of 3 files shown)


ayan-usgs:
:+1:

Merge at your discretion.

As a general question: How is `updateNldiSites` function on https://github.com/mbucknell/WQP_UI/blob/d2908c9982b70108d1947d90436434ca0ec6b729/portal_ui/static/js/views/nldiView.js#L138 being called? More curious than anything else.

mbucknell:
FYI: I have a few more changes to make to handle large number of sites. Going to add Leaflet.Cluster! Anyways, updateNldiSites is now poorly named. It should be named something like updateNldiParameter. The idea is that you would call this whenever your comid, navigation, or direction is updated. So it is called from the click mouse event handler, findSitesHandler, because that is the handler to find the comid that you clicked on, and it is called from the change event handler for the navigation select and from the change event handler for distance text field. Both of these are within the nldiControl so those event handlers are passed into the function which creates the controls .

ayan-usgs:
Cool.

Yay Leaflet Cluster!

mbucknell:
The other thing I should mention is the "addSingleClickHandler" is provided by MapWithSingleClickHandler defined as follows: 

var MapWithSingleClickHandler = L.Map.extend({
		includes : L.SingleClickEventMixin
	});

The SingleClickEventMixin is defined in LeafletMixins/SingleClickEventMixin.js

mbucknell:
Going to merge this pull request. Additional changes to handle large number of sites will be done in another pull request.

PR #123, opened 2016-07-01
Wqp 823

mbucknell:
Rather than having the user click on a comid, we are now displaying a layer that shows pour points. The user clicks on one of these, its corresponding huc12 is retrieved and then used when retrieving the sites and flow lines and when making queries.

### DIFF
--- a/portal_ui/templates/portal.html
+++ b/portal_ui/templates/portal.html
@@ -325,7 +325,14 @@
 			</div>
 			<div class="dummy-row"></div>
 			<div class="clearfix"></div>
-			<div id="nldi-map"></div>
+			<div id="nldi-map-container">
+				<div id="nldi-help-div">
+					Zoom in to see the pour point layer. Once a navigation type
+					and optionally a distance are specified, click on a pour point to select
+					WQP sites for your query.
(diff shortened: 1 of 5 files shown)


portal_ui/templates/portal.html
```diff
@@ -325,7 +325,14 @@
 			</div>
 			<div class="dummy-row"></div>
 			<div class="clearfix"></div>
-			<div id="nldi-map"></div>
+			<div id="nldi-map-container">
+				<div id="nldi-help-div">
+					Zoom in to see the pour point layer. Once a navigation type
+					and optionally a distance are specified, click on a pour point to select
```
ayan-usgs: As nerdy as this sounds, I just spent the last five minutes trying to decide if the subject-verb agreement is kosher here. I think it is, though maybe `... an optional distance...` is a bit more clear.

ayan-usgs:
:+1: 

Looks good to go!

portal_ui/templates/portal.html
mbucknell: Jim will change it anyways so I'm going to leave as is.

PR #125, opened 2016-07-20
Celery

jkreft-usgs:
updating to asynchronous for cache management, also kicking up some version numbers, including for flask proper


portal_ui/tasks.py
```diff
@@ -0,0 +1,66 @@
+from celery import Celery
+from . import app
+import requests
+import redis
+import csv
+import cPickle as pickle
```
mbucknell: Order of imports should  be as suggested in https://www.python.org/dev/peps/pep-0008/#imports .

jkreft-usgs: I had no idea of that bit of PEP8

portal_ui/tasks.py
```diff
@@ -0,0 +1,66 @@
...
+
+celery = Celery(app.name, broker=app.config['CELERY_BROKER_URL'])
+celery.conf.update(app.config)
+
+
+@celery.task(bind=True)
+def generate_site_list_from_streamed_tsv_async(self, base_url, redis_config, provider_id, redis_db):
```
mbucknell: What does self refer to? This is typically used within a class definition but a very quick look at the Celery docs don't make it clear what self would refer to.

portal_ui/tasks.py
```diff
@@ -0,0 +1,66 @@
...
+                header_line = line
+                header_object = csv.reader([header_line], delimiter='\t')
+                for row in header_object:
+                    header = row
+                counter += 1
+            elif counter >= 1:
+                station = csv.reader([line], delimiter='\t')
+                for row in station:
+                    station_data = row
+                    if len(station_data) == 36:
```
mbucknell: Where does 36 come from? Consider making this a named constant.

portal_ui/tasks.py
jkreft-usgs: it allows for updating the state of the task. (basically, the task decorator wraps the whole thing in a class.)  This gives a bit more detail, and actually taking this approach with a complex task like this might be the way to make it testable in the future. http://jsatt.com/blog/class-based-celery-tasks/

portal_ui/templates/cache_manager.html
```diff
@@ -0,0 +1,92 @@
...
+                    // rerun in 2 seconds
+                    setTimeout(function() {
+                        update_progress(status_url, nanobar, status_div);
+                    }, 2000);
+                }
+            });
+        }
+
+        function on_submit_action()
+    {
```
mbucknell: OCD alert: Indentation

portal_ui/tasks.py
mbucknell: But you are not wrapping it in a class? I can see how doing so might make testing easier though. It would be something that could be mocked out.

mbucknell:
In general it looks good to go 👍

portal_ui/tasks.py
jkreft-usgs: 36 is the number of columns I expect (basically, I want the number of columns in the row to matach the number of columns in the header).  I can make this much more robust to adding columns from the WQP by changing the hard-coded number to a variable.

jkreft-usgs:
I'm going to leave the cache manager page alone for now, since we all know it needs to be rebuilt if it is going to actually be pretty.

PR #128, opened 2016-08-04
Fix bower

mbucknell:


mbucknell:
@ayan-usgs , going to commit and find out what happens on build. Let me know if you have issues locally, but I'm hoping you don't have to specify anything else on the command line when installing bower dependancies.

PR #130, opened 2016-08-04
Wqp 809

ayan-usgs:


portal_ui/static/js/ol3Utils/mapUtils.js
```diff
@@ -35,6 +35,22 @@ WQP.ol3.mapUtils = (function() {
...
+		var esriHydroURL = 'http://hydrology.esri.com/arcgis/rest/services/WorldHydroReferenceOverlay/MapServer/tile/';
+		var hydroTileLayer = new ol.layer.Tile({
+			title : 'ESRI Hydro Layer',
+			map : map,
+			visible : isVisible,
+			source : new ol.source.XYZ({
+				url : esriHydroURL + '{z}/{y}/{x}'
+			})
+		});
+		layerDeferred.resolve(hydroTileLayer);
```
mbucknell: Can you explain why you are returning a promise? Seems like the tile layer creation is not asychronous.

ayan-usgs: I wanted to return a chainable object, because it nicely encapsulated the Z-Index setting and the addition of the layer to the overlay group. But, now that you mention it, that is superfluous. I'll refactor.

mbucknell: This makes me think that we should probably be setting the zIndex for the NWISSitesLayer as well. I think we are just getting lucky that it's ended up on top. It would be simple to add this.

ayan-usgs: I thought about that, but didn't because it seemed to be working. What's strange is that the NWISSitesLayer has a z-index of 0 (same as the base layers)... you're probably right that we're getting lucky.

I'll make the addition.

ayan-usgs:
@mbucknell,

All changes have been made if you want to take a look.

portal_ui/static/js/map/siteMap.js
```diff
@@ -76,11 +102,19 @@ PORTAL.MAP.siteMap = function(options) {
...
+		});
+		pushLayer(esriHydroLayer, overlayLayerGroup);
+
+		// add the NWIS Sites Layer
 		WQP.ol3.mapUtils.getNWISSitesLayer({}, {
 			visible : false,
 			map : map
 		}).done(function(layer) {
-			overlayLayerGroup.getLayers().push(layer);
+			pushLayer(layer, overlayLayerGroup);
```
mbucknell: I wonder if it it makes sense to figure out what the max zIndex for the base layers are and then set the zIndex appropriately for the overlay layers (i.e. add 1 for the Hydro layers, add 2 for the NWISSiteLayer)? I think that would give the coder control over the layer stacking.

mbucknell:
See new comment. Good to go. 👍

portal_ui/static/js/map/siteMap.js
ayan-usgs: @mbucknell:

I think that's a good idea. It's a quick change in the `pushLayer` function where it does `setZIndex()` call.

PR #138, opened 2016-09-20
Wqp 896

ayan-usgs:


README.md
```diff
@@ -40,19 +40,21 @@ REDIS_CONFIG = None
...
 ```
 
-
+## Setup on Linux or MacOS
 To build this application on a linux or Mac OS you can use the dev_install.sh script. If you want a fresh install type in:
-`source dev_install.sh --clean`
-This will remove the previous javascript and python dependencies. If you don't want to install from scratch, type in:
 `source dev_install.sh`
+This will remove the previous javascript and python dependencies. If you don't want to install from scratch, type in:
+`source dev_install.sh --update`
```
mbucknell: Why did you change the default from no cleaning? I was following the pattern of maven where you have to specify the clean goal on the command line. Note that all clean does is remove directories. If dependancy versions change, pip will still update them to match what is in requirements.txt. Ditto for bower.

README.md
```diff
@@ -40,19 +40,21 @@ REDIS_CONFIG = None
...
-4. Run `pip install -r requirements.txt`
+## Setup on Windows
+To build on Windows, you can use the dev_install.ps1 script. If you want a fresh install type in:
+`.\dev_install.ps1`
+This will remove previous javascript and python dependencies. If you don't want to install from scratch, type in:
+`.\dev_install.ps1 --update`
+This will update javascript and python dependencies. Both commands will run the jasmine tests on Windows.
 
+## Manual Setup
 If you don't have maven installed, you can do the following steps:
```
mbucknell: The project doesn't need maven at all anymore, so statement should be removed.

dev_install.ps1
```diff
@@ -0,0 +1,36 @@
+param([String]$update="")
+
+Write-Output "You must have node, karma, Python 2.7, and virtualenv installed and available via the system path."
+
+$configExists = Test-Path instance/config.py
+if (-Not $configExists) {
+	Write-Output "Please create an instance/config.py file before proceeding. See the README.md for what's required."
+	Exit
+}
+
```
mbucknell: Cleaning should delete the contents of the node_modules and bower_components directories. As I said below, I believe install will only update the dependancies that have changed. It should also remove the virtualenv, env, and the static/gen/ and static/.webassets-cache for the wqp project.

dev_install.ps1
```diff
@@ -0,0 +1,36 @@
...
+	bower update
+}
+else {
+	Write-Output "Installing npm and bower dependencies"
+	npm install
+	bower install
+}
+
+Write-Output "Running Javascript tests."
+karma start test/js/karma.conf.js
```
mbucknell: Use the path name to the node installed karma.

dev_install.sh
```diff
@@ -2,20 +2,28 @@
```
mbucknell: Above comments for the windows version applies here as well

dev_install.ps1
ayan-usgs: I would strongly recommend not specifying the path the node installed karma. The node installed karma is inoperable on Windows... it installs as a shell script that is unexecutable by Windows. 

The karma.sh included in node_modules looks like this:
```
#!/usr/bin/env node

require('../lib/cli').run();
```
That does nothing for windows.

It's more effective to install karma globally and set the path to karma.cmd. I'm not hugely inclined to have the script go search for it... there are too many variables such as path and user permissions to get caught up in.

README.md
ayan-usgs: I wasn't defining "clean" in the maven sense. I was thinking of it more in-terms of a first install after a clone, not removing any directories because there aren't any directories to be removed. 

If the issue is with the text, that can be changed.

I actually don't find the argument of following maven patterns to be hugely compelling because this isn't a maven project anymore. Making it look like a maven project strikes me as window dressing.

dev_install.ps1
ayan-usgs: Huh.. what do you know... that using the one in node_modules works. I recall that not working on windows a couple months ago....

README.md
mbucknell: It's not so much following the maven pattern IMHO. It's what do we want clean to do. One reason to do a clean is to make sure the requirements.txt, bower.json, and package.json are up to date. For example let's say I have installed a new javascript library but forgot to put it in bower.json. Doing a clean build prior to putting in a pull request should catch this mistake.

Since cleaning is a more destructive process, forcing the user to specifically ask for it makes sense to me.

dev_install.ps1
ayan-usgs: I actually turns out that powershell shell can't use karma.sh... it only worked because I used git bash last time. :anguished:

PR #140, opened 2016-09-26
Wqp 412

mbucknell:
Made the web app a blueprint in preparation for adding a sites blueprint which will provide a sites service. It may make sense in the future to break portions of the webapp into their own blueprints. I'm thinking of things like the geoserver proxies, and things like the cachemanagement page. But for now they are in the portal_ui blueprint.

This is only the first step in this ticket. Because of the large number of files and the organizational changes I felt an initial pull request was in order.

Also updated the scripts to consolidate the cleaning task and to add the static/gen and static/.webassets-cache directoreis that are generated.


mbucknell:
@ayan-usgs and @jkreft-usgs , feel free to comment, but most of the changes were needed to move the portal_ui into a blueprint.

ayan-usgs:
The new divisional structure looks fine. There are some implications for the build in terms of what get's packaged in the artifact and what's in `MANIFEST.in`, but those should be straight-forward changes.

mbucknell:
@ayan-usgs , is there a way to package the artifact locally so that a developer can test that everything is updated correctly. Maybe there needs to be a package.sh?

ayan-usgs:
@mbucknell,

To build an artifact locally, I'd recommend running `python setup.py sdist` or `python setup.py bdist_wheel`. The `wheel` package will need to be installed for the later. Either way, the output from the build will let tell you if it's packaged everything correctly. That having been said, a source distribution is particularly useful for a human to inspect because it's just a zip file, so I'd suggest that rather than a wheel.

PR #148, opened 2016-09-28
Wqp 412

mbucknell:
Removed the debug and logging there were added. Fixed the install scripts and added ASSETS_AUTO_BUILD set to False in config.py


mbucknell:
@ayan-usgs , please merge this when you are ready.

PR #149, opened 2016-10-03
Wqp 412

mbucknell:
Added the sites blueprint. Using streaming and generators to provide a streaming nwis site service that provides a geojson feature collection containing the sites requested.

I would like to add view tests in addition to the utils tests. If I have time before the next sprint, I am try to add them.


wqp/sites/utils.py
```diff
@@ -0,0 +1,194 @@
...
+
+def site_feature_generator(iter_lines):
+    """
+    Generator which yields a geosjon Feature object for each line representing a station in rdb_text
+
+    :param iter_lines: Generator which yields a line. The lines are assumed to represent an rdb formatted NWIS
+        site file.
+    :yield: geojson Feature object for each station in rdb_text
+    """
+    print('Start feature generation')
```
ayan-usgs: I've been trying to move away from using `print` functions in production code. I prefer logging statements because they're more information dense.

wqp/sites/tests/test_utils.py
```diff
@@ -0,0 +1,140 @@
+
+from unittest import TestCase
+
+from geojson import loads as geojson_loads, dumps as geojson_dumps
+from requests_mock import Mocker as Requests_Mocker, ANY
```
ayan-usgs: How is `ANY` being used? I'm don't see it being used in any `register_uri` methodcalls.

ayan-usgs: Good to go. Nothing major found.

mbucknell: It isn't, I used it and then forgot to take it back out.

wqp/sites/utils.py
mbucknell: Agreed. I had this in there for debuggin and will remove it.

PR #157, opened 2016-10-06
The git+ requirement has to be specified as a dependency_link in setu…

mbucknell:
…p. Modified the script to find requirements that begin with git+.


mbucknell:
@ayan-usgs , Feel free to comment if you see a better way of doing this. I'm going to merge to see if it works in our build process.

PR #170, opened 2016-10-11
some tidying-up in setup.py

ayan-usgs:


ayan-usgs:
Yeah. Missed that. I'll get rid of it.

PR #181, opened 2016-10-19
Wqp 799

ayan-usgs:
Three new templates and a lot of .pngs.


wqp/wqx/templates/wqx/index.html
```diff
@@ -0,0 +1,13 @@
+<html>
+<head>
+	<title>WQX-Outbound Schema Info</title>	<style type="text/css" media="screen">		.instructions {			text-align: center;			border-width: 2pt;			border-color: #333333;			border-style: solid;			background-color: #FFEE99;			padding-top: 4pt;			padding-bottom: 2pt;			padding-left: 8pt;			padding-rightt: 8pt;			margin-bottom: 8pt;		}		.note {			font-style: italic;			font-size: small;		}		body {			font-family: arial;		}	</style>
```
mbucknell: This seems a little wonky. I don't see the ending </title> tag or the beginning <script> tag.

wqp/wqx/templates/wqx/index.html
```diff
@@ -0,0 +1,13 @@
...
+</head>
+<body>	<div class="instructions">The National Water Information System (NWIS) Water-Quality 	Web Services conform to the format defined in the below referenced XML schema.	</div>	<b>WQX-Outbound XML schema information:</b>
```
mbucknell: This also seems wonky with the multiple spaces between "Water-Quality" and "Web Services"

ayan-usgs: That's what I get for using the existing index.html verbatim. :gun:

The weird tab has been removed.

wqp/wqx/templates/wqx/index.html
ayan-usgs: Which html tag isn't present?

mbucknell: They are all there. My mistake but I don't like putting the style on the same line as the title tag. It's just formatting but it makes it hard to read.

PR #183, opened 2016-10-20
run python and karma tests in travis

ayan-usgs:


.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+  - npm install
+  - node_modules/bower/bin/bower install
+
+before_script:
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
+  - echo "SECRET_KEY = 'key-for-testing'" > instance/config.py
+
+script:
+  - nosetests
```
ayan-usgs: I would have used `python setup.py test` but there were versioning problems. Specifically it kept getting kill by `kombu` not being happy with the version of `amqp`; `kombu` wants a version of `amqp` that is less than 2.0.X. I wasn't too inclined to do a downgrade with the requirements... particularly when things as a whole are working with the app.

.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+  - node_modules/karma/bin/karma start test/js/karma.conf.js --no-colors --log-level debug
```
mbucknell: Don't think we need to specify a log-level of debug during routine tests

PR #190, opened 2016-10-26
Wqp 929

mbucknell:
Created a new leaflet control, QuerySelectControl, which creates a pick list for the feature source to query on. This will pick a layer to show when selected.

Refactored the nldiView to move the data into nldiModel. This should make adding new query sources simpler. The constant QUERY_SOURCES just needs to be updated with a new element containing the required information. This assumes that the WMS GetFeatureInfo will be used to query for the site at a lat lon clicked.


test/js/portal_ui/nldiModelSpec.js
```diff
@@ -0,0 +1,88 @@
...
+
+	it('Expects that reset returns the model to it\'s initial state', function() {
+		var data;
+		nldiModel.setData('featureId', '1234');
+		nldiModel.setData('navigation', nldiModel.NAVIGATION_MODES[1]);
+		nldiModel.setData('newProp', 'this');
+		nldiModel.reset();
+		data = nldiModel.getData();
+
+		expect(_.has(data, 'featureSource')).toBe(true);
```
ayan-usgs: I'm curious what advantage `toBe(true)` has over `toBeTruthy()`. It seems a bit inconsistent to use `toBe(true)` in this block and `toBeFalsy()` in the subsequent block.

mbucknell: The test toBe(true) is more strict than toBeTruthy(). In the former case the resut being tested must be a boolean value so it has to be true. In the later case, it just has to be truthy so a non empty string, non empty array, or true all would pass the test. I have to use toBeFalsy in the subsequent block because the result is not necessarily a boolean.

PR #194, opened 2016-11-16
do not include the .webassets-cache directory in the build

ayan-usgs:


ayan-usgs:
@mbucknell,

Yeah, I tried to find a general way to just say ignore anything that fits the pattern *.webassets-caches*, but didn't have much luck. Ended up having to explicitly exclude the directory both in the manifest and in setup.py when it's sleuthing around for data files.

PR #195, opened 2016-11-16
switch URLs in config.py from http to https

ayan-usgs:


jkreft-usgs:
We can't actually do this bit yet- the nwis web based services don't support https yet- but fortunately, we don't call anything for them clientside, so this can wait.

ayan-usgs:
Alright. I'll close this PR.

PR #197, opened 2016-11-17
use sessions to persist SSL cert verification across multiple requests

ayan-usgs:


README.md
```diff
@@ -11,6 +11,7 @@ You will need to create a config.py file in the `instance` directory. It should
 ```python
 
 DEBUG = True
+VERIFY_CERT = True
```
mbucknell: So we want to set VERIFY_CERT to true locally. I thought in pubs we are setting it to False for local development?

PR #198, opened 2016-11-17
change more links from http to https

ayan-usgs:


wqp/portal_ui/static/js/views/nldiView.js
```diff
@@ -323,9 +323,9 @@ PORTAL.VIEWS.nldiView  = function(options) {
 		'World Imagery' : L.tileLayer.provider('Esri.WorldImagery')
 	};
 	var hydroLayer = L.esri.tiledMapLayer({
-		url : "http://hydrology.esri.com/arcgis/rest/services/WorldHydroReferenceOverlay/MapServer"
+		url : "https://hydrology.esri.com/arcgis/rest/services/WorldHydroReferenceOverlay/MapServer"
```
mbucknell: I'm getting a time out when I changed this to https locally. http works fine. This could be a problem

ayan-usgs: hmm... yeah...

ayan-usgs:
> Just FYI: Links don't have to change to https, only ajax calls are subject to the mixed content warnings.

I know, but figure might as well use the https URL. They'll probably get redirected from http to https anyway.

wqp/portal_ui/static/js/views/nldiView.js
mbucknell: We may have to explore proxying this particular layer. If we do that I believe it's a new ticket.

PR #200, opened 2016-11-21
Wqp 938

mbucknell:
Using Leaflet 1.0.x rather than OpenLayers 3 to implement the show sites on map feature. Note that that identify sites feature is not part of this ticket. It should be the only feature not reimplemented.


wqp/portal_ui/static/js/views/nldiView.js
```diff
@@ -323,7 +323,7 @@ PORTAL.VIEWS.nldiView  = function(options) {
 		'World Imagery' : L.tileLayer.provider('Esri.WorldImagery')
 	};
 	var hydroLayer = L.esri.tiledMapLayer({
-		url : "http://hydrology.esri.com/arcgis/rest/services/WorldHydroReferenceOverlay/MapServer"
+		url : Config.HYDRO_LAYER_ENDPOINT
 	});
 	var nhdlPlusFlowlineLayer = L.tileLayer.wms('https://cida.usgs.gov/nwc/geoserver/gwc/service/wms',
```
ayan-usgs: A bit out of the purview of this task, but I think it would be good to eventually move this to the configuration as you did with the hydro layer.

dev_install.sh
```diff
@@ -21,25 +21,25 @@ fi
...
-node_modules/karma/bin/karma start test/js/karma.conf.js;
-
 if [ ! -s env ]; then
     echo "Creating the virtualenv env";
 	virtualenv --python=python2.7 env;
 fi
 echo "Installing python requirements";
 env/bin/pip install -r requirements.txt;
 
+echo "Running Javascript tests";
```
ayan-usgs: Should the powershell variant of this script get similar changes too?

wqp/portal_ui/static/js/views/nldiView.js
mbucknell: I think I'll do it. Good to get these out into config variables.

dev_install.sh
mbucknell: Probably should.

PR #201, opened 2016-11-28
Wqp 939

mbucknell:
Added in a single click handler to handle the click-to-identify features. Added Leaflet.Draw to use the draw tools to draw a rectangle for the box-identify feature. The GetFeature request now uses the application/json mimeType which caused minor changes in the json passed to the the identifyDialog object.

### DIFF
--- a/wqp/portal_ui/static/js/leafletLayers/WQPSitesLayer.js
+++ b/wqp/portal_ui/static/js/leafletLayers/WQPSitesLayer.js
@@ -83,6 +83,13 @@
 			return Config.SITES_GEOSERVER_ENDPOINT + 'wms?' + $.param(queryParams);
 		},
 
+		/*
+		 * @param {L.LatLngBounds} bounds
+		 * @returns {Jquery.Promise}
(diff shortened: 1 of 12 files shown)


wqp/portal_ui/static/js/leafletLayers/WQPSitesLayer.js
```diff
@@ -88,11 +109,12 @@
...
 				service : 'wfs',
 				version : WFS_VERSION,
-				typeName : LAYER_NAME,
+				typeNames : LAYER_NAME,
 				SEARCHPARAMS : getSearchParams(queryParamArray),
 				outputFormat : 'application/json'
 			};
-			return Config.SITES_GEOSERVER_ENDPOINT + 'wfs/?' + $.param(queryData);
+
+			return Config.SITES_GEOSERVER_ENDPOINT  + 'wfs/?' + $.param(queryData);
```
ayan-usgs: This is nit-picky thing that doesn't matter in the scheme of things, but it doesn't look like the extra a space preceding the `+` is intentional.

PR #202, opened 2016-11-28
force SSL

jkreft-usgs:
Force SSL if it isn't already happening


wqp/__init__.py
```diff
@@ -5,15 +5,15 @@
...
+from flask_sslify import SSLify
 
 __version__ = '4.3.0dev'
 
 
 app = Flask(__name__.split()[0], instance_relative_config=True)
 
 Bower(app)
-
+sslify = SSLify(app, permanent=True)
```
mbucknell: I would put this after loading the config as it's behavior is affected by config variables.

jkreft-usgs: it seems to work fine right now, in that if i change to ``DEBUG=True`` in my instance config, it starts setting the SSL properly

PR #210, opened 2016-12-09
Wqp 940

mbucknell:


wqp/portal_ui/templates/coverage.html
```diff
@@ -3,107 +3,79 @@
 {% block title %}WQP National Results Coverage{% endblock %}
```
mbucknell: Changes to this where primarily formatting changes with a few changes to the layer parameter names and eliminating the openlayer script loading.

wqp/portal_ui/static/js/coverage/coverageMap.js
```diff
@@ -0,0 +1,132 @@
...
+/* global Handlebars */
+/* global _ */
+/* global $ */
+
+var COVERAGE = COVERAGE || {};
+
+/*
+ * @param {Object} options
+ * 		@prop {String} mapDivId - Div where map will be rendered
+ * 	    @prop {Jquery element} $legendImg - image element which will contain the legend
```
ayan-usgs: A bit nit-picky, but the indents in these doc strings look a bit off.

wqp/portal_ui/static/js/leafletLayers/CoverageLayer.js
```diff
@@ -0,0 +1,141 @@
...
+		'all': 'A'
+	};
+	var SLD_FEATURE = {
+		'states': 'S',
+		'counties': 'C',
+		'huc8': 'H'
+	};
+	var SLD_TIMEFRAME = {
+		'past_12_months': '1',
+		'past_60_months': '5',
```
ayan-usgs: I'm wondering if it would be more readable to call these parameters `past_1_year` and `past_5_years` or something. Not a bit deal either way.

mbucknell: Have to leave those strings as is. That is what is passed in the VIEWPARAMS when creating the layer and when retrieving information using GetFeature

PR #214, opened 2016-12-16
Wqp 760

mbucknell:


wqp/portal_ui/static/js/downloadProgressDialog.js
```diff
@@ -1,111 +1,122 @@
...
+			title: 'Download Status',
+			continueMessage: 'download the data',
+			cancelDownload: function (counts, fileFormat) {
+				return (counts !== 0) && (fileFormat === 'xlsx') && (parseInt(counts.split(',').join('')) > 1048575);
+			},
+			cancelMessage: 'Your query is returning more than 1,048,575 results which exceeds Excel\'s limit.'
+		}
+	};
+
+	var opKind;
```
ayan-usgs: Kind of nit-picky, but would I'd prefer the variable to defined a bit closer to where it's used. `opKind` doesn't seem to be used until L60.

mbucknell: I think also a comment about it's use is warrented. I'll add that.

PR #216, opened 2016-12-23
Wqp 937

ayan-usgs:


wqp/portal_ui/static/js/nldiMapping.js
```diff
@@ -0,0 +1,147 @@
...
+	var upstreamLineStyle = {
+		"color": "#253494",
+		"weight": 5,
+		"opacity": 0.65,
+		"dashArray": '15,8',
+		"lineJoin": 'square'
+	};
+
+	var onEachPointFeature = function(feature, layer) {
+		var parser = document.createElement('a');
```
mbucknell: What is the purpose of creating parser? It looks like it is created and then the pathname is extracted. I think you can use feature.properties.uri.

wqp/portal_ui/static/js/nldiMapping.js
```diff
@@ -0,0 +1,147 @@
...
+/* global L */
+/* global Config */
+/* global _ */
+/* global $ */
+
+
+var NLDI = NLDI || {};
+
+
+NLDI.overlays = function(map) {
```
mbucknell: Since this method does not return anything it would be more appropriate to use a verb rather than a noun as the method name. Perhaps something like addOverlays. This function should have jsdocs comments.

wqp/portal_ui/static/js/nldiMapping.js
```diff
@@ -0,0 +1,147 @@
...
+
+	var upstreamLineStyle = {
+		"color": "#253494",
+		"weight": 5,
+		"opacity": 0.65,
+		"dashArray": '15,8',
+		"lineJoin": 'square'
+	};
+
+	var onEachPointFeature = function(feature, layer) {
```
mbucknell: A more descriptive name would make the purpose of this method clearer. Something like addPopupToPointFeature.

wqp/portal_ui/static/js/nldiMapping.js
```diff
@@ -0,0 +1,147 @@
...
+			onEachFeature: onEachPointFeature,
+			pointToLayer: function (feature, latlng) {
+				return L.circleMarker(latlng, markerOptions);
+			}
+		});
+		markers.addLayer(pointLayer);
+		map.addLayer(markers);
+	};
+
+	var onEachLineFeature = function(feature, layer) {
```
mbucknell: Consider a more descriptive method name.

wqp/portal_ui/static/js/nldiMapping.js
```diff
@@ -0,0 +1,147 @@
...
+		layer.bindPopup(popupText);
+	};
+
+	var addLineDataToMap = function(data, style) {
+		var lineLayer = L.geoJson(data, {
+			onEachFeature: onEachLineFeature,
+			style: style
+			});
+		lineLayer.addTo(map);
+		_.each(data.features, function(feature) {
```
mbucknell: I'd like to see us start using the native javascript forEach (http://www.w3schools.com/jsref/jsref_forEach.asp)  and map (http://www.w3schools.com/jsref/jsref_map.asp) rather than underscore's _.each and _.map Particularly in Leaflet files and particularly if it eliminates the dependency on underscore. 

For this particular function map would be more appropriate because you are basically copying the feature to allExtents.features. Also not sure why the context is being passed into _.each.

wqp/portal_ui/static/js/nldiMapping.js
```diff
@@ -0,0 +1,147 @@
...
+		});
+	};
+
+	var addNldiPointsToMap = function(endpointUrl, style) {
+		$.getJSON(endpointUrl, distanceParam, function(data) {
+			addPointDataToMap(data, style);
+		});
+	};
+
+	var wqpUrlUt = nldiUrl + f + "/" + c + "/navigate/" + e + "/wqp";
```
mbucknell: Consider making these names constants using uppercase.

wqp/portal_ui/static/js/nldiMapping.js
```diff
@@ -0,0 +1,147 @@
...
+	var wqpUrlDm = nldiUrl + f + "/" + c + "/navigate/" + g + "/wqp";
+	var nhdUrlUt = nldiUrl + f + "/" + c + "/navigate/" + e;
+	var nhdUrlDm = nldiUrl + f + "/" + c + "/navigate/" + g;
+	var wqpUrlSite = nldiUrl + f + "/" + c + "/";
+
+	// style the current site so it can be easily identified
+	$.getJSON(wqpUrlSite, {}, function(data) {
```
mbucknell: Consider moving this below the variable declarations for nldiLines and nldiPoints so all executable code is together.

wqp/portal_ui/static/js/providerSiteMap.js
```diff
@@ -0,0 +1,32 @@
...
+
+SITE.siteMap = function(options) {
+	"use strict";
+	var mapDivId = options.mapDivId;
+	var map;
+	var site = Config.site;
+	map = MAPS.create(mapDivId, 'Esri.WorldGrayCanvas');
+
+	var esriHydroLayer = L.esri.tiledMapLayer({
+		url: "http://hydrology.esri.com/arcgis/rest/services/WorldHydroReferenceOverlay/MapServer"
```
mbucknell: This endpoint should be retrieved from the app configuration (HYDRO_LAYER_ENDPOINT)

wqp/portal_ui/static/js/providerSiteMap.js
```diff
@@ -0,0 +1,32 @@
...
+	});
+	esriHydroLayer.addTo(map);
+
+	var nhdplusLayer = L.tileLayer.wms('https://cida.usgs.gov/nwc/geoserver/gwc/service/wms', {
```
mbucknell: This endpoint should be retrieved from the app configuration(NHDPLUS_FLOWLINE_ENDPOINT) and the layer name from NHDPLUS_FLOWLINE_LAYER_NAME

wqp/portal_ui/static/js/providerSiteMap.js
```diff
@@ -0,0 +1,32 @@
+/* jslint browser: true */
+/* global L */
+/* global Config */
+/* global _ */
+/* global $ */
+
+var SITE = SITE || {};
+
+
+SITE.siteMap = function(options) {
```
mbucknell: Please add jsdoc comments. Also consider passing in the initial lat/lon and zoom since this is different for the the two maps.

wqp/portal_ui/static/js/providerSitesMap.js
```diff
@@ -0,0 +1,76 @@
...
+/* global _ */
+/* global $ */
+
+
+var SITES = SITES || {};
+
+
+SITES.sitesMap = function(options) {
+	"use strict";
+	var self = {};
```
mbucknell: I don't think you need this. This method doesn't return anything. Please add doc comments to the sitesMap method and change it to a verb (maybe createSitesMap or initializeSitesMap)

wqp/portal_ui/static/js/providerSitesMap.js
```diff
@@ -0,0 +1,76 @@
...
+					fillOpacity: 0.7
+				});
+			}
+		});
+		markers.addLayer(dataLayer);
+		map.addLayer(markers);
+		map.fitBounds(L.geoJson(data).getBounds());
+	};
+
+	self.sitesMap = addDataToMap(siteData);
```
mbucknell: addDataToMap doesn't return anything. I think you can just call it.

wqp/portal_ui/templates/sites.html
```diff
@@ -35,82 +35,13 @@
     <script type="text/javascript" src="{{ url_for('bower.static', filename='leaflet/dist/leaflet.js') }}"></script>
 	<script type="text/javascript" src="{{ url_for('bower.static', filename='leaflet-providers/leaflet-providers.js') }}"></script>
     <script type="text/javascript" src="{{ url_for('bower.static', filename='leaflet.markercluster/dist/leaflet.markercluster.js') }}"></script>
-    <script type="text/javascript">
+	<script type="text/javascript">
+		Config.siteData = {{ sites_geojson|tojson }};
```
mbucknell: This really picky, but can we call this sitesData since it's multiple sites.

wqp/portal_ui/static/js/nldiMapping.js
jkreft-usgs: I like just using the URI as well, because it means that this code isn't tightly bound to just dealing with WQP data coming from the NLDI.  It does mean that we will be linking to prod from dev(not ideal), but we should be making the change on the service side to fix that issue, not parsing the content on the client.

ayan-usgs: I was pretty much trying to avoid linking out of the tier. In addition, the parser was in the original code, so the implication seemed that someone had previously made the decision the deal with parsing on the client rather than in the services -- it didn't seem necessary to countermand a decision on priority of those two items when one had already been made.

jkreft-usgs: The previous decision made to parse on the client was made by me, though I had forgotten doing that (it definitely makes local development easier). Clearly, there are reasons to do it both ways.  Linking out of the tier isn't great, but neither is doing a bunch of parsing on production where the data coming out of the NLDI is already correct.  This is also a challenge since NLDI is actually a separate system, which should be pulling from the production WQP data, and we should be accessing the production NLDI on dev WQP.  In conclusion, how to link and interact with OWI-managed outside systems is a land of contrasts. :/

mbucknell: It seems to me we should be pulling the url from the data that is returned. This is straightforward. It's a link then to the production site, but I think we can live with that. Keep it simple

ayan-usgs: That's fine with me. :smile:

ayan-usgs:
Addressed review comments.

PR #217, opened 2016-12-23
get around broken url mapping

jkreft-usgs:
take advantage of our good friend LOCAL_BASE_URL to get around our kinda frustrating url mapping.

### DIFF
--- a/wqp/portal_ui/views.py
+++ b/wqp/portal_ui/views.py
@@ -374,7 +374,7 @@ def sitescachetask(provider_id):
     redis_db = generate_redis_db_number(provider_id)
     task = generate_site_list_from_streamed_tsv_async.apply_async(args=[base_url, redis_config,
                                                                           provider_id, redis_db])
-    response_content = {'Location': url_for('portal_ui.taskstatus', task_id=task.id)}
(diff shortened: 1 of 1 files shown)


wqp/portal_ui/views.py
```diff
@@ -374,7 +374,7 @@ def sitescachetask(provider_id):
     redis_db = generate_redis_db_number(provider_id)
     task = generate_site_list_from_streamed_tsv_async.apply_async(args=[base_url, redis_config,
                                                                           provider_id, redis_db])
-    response_content = {'Location': url_for('portal_ui.taskstatus', task_id=task.id)}
+    response_content = {'Location': app.config['LOCAL_BASE_URL']+"/status/"+task.id}
```
ayan-usgs: Minor detail, but I'd prefer spaces between operators per https://www.python.org/dev/peps/pep-0008/#other-recommendations.

jkreft-usgs: just swapped to .join.  I've been on a kick of avoiding string concatenation lately anyways... and apparently it is even recommended: https://www.python.org/dev/peps/pep-0008/#id49

PR #219, opened 2017-01-19
Wqp 925

mbucknell:


wqp/portal_ui/static/js/portalHelp.js
```diff
@@ -139,6 +139,13 @@ PORTAL.MODELS.help = {
                 </ul>\
                 </div>'
 	},
+	minactivities: {
+		placement: 'auto',
+		title: 'Minumum Sampling Activities Per Site Help',
+		content: '\
+			<div>This limits the data returned to data from sites where at least a minimum number of sampling activities have been \
```
ayan-usgs: This seems a bit confusing... would this be a bit clearer?

"This limits returned data to those sites with reported activities greater or equal to the minimum number that also conform to the other query parameters."

mbucknell: Not going to change this as Jim will likely update.

PR #227, opened 2017-02-02
Wqp 1008 - Upstream-downstream Query Map support

mbucknell:
The workflow for the NLDI upstream/downstream query is now: user clicks the expand button on the inset map, zooms in and clicks the feature that they want to navigate from. This pops up a dialog which provides meta data about the chosen feature and allow the user to pick a navigation type and distance and then perform the navigation query.


test/js/karma.conf.js
```diff
@@ -38,6 +38,7 @@ module.exports = function(config) {
         'wqp/bower_components/Leaflet.EasyButton/src/easy-button.js',
         'test/resources/testConfig.js',
         'wqp/portal_ui/static/js/**/*.js',
+        {pattern: 'wqp/portal_ui/static/js/hbTemplates/*.hbs', included: false},
```
mbucknell: You have to have include set to false so that karma doesn't read these in at initialization. The templates are read in by $.ajax within the code.

test/js/portal_ui/leafletControls/FeatureSourceSelectControlSpec.js
```diff
@@ -0,0 +1,151 @@
+/* jslint browser: true */
```
mbucknell: Refactored QuerySelectControl to FeatureSourceSelectControl since that is the terminology that we use. I did add the ability to set an initial value for this control.

PR #229, opened 2017-02-07
Wqp 1027 - Give hints to ArcGIS Online users

mbucknell:


wqp/portal_ui/static/js/hbTemplates/arcGisHelp.hbs
```diff
@@ -0,0 +1,18 @@
+<div>
+	<p> Please copy and paste into the appropriate parameter fields in ArcGIS Online
+		when adding a layer from the web.
```
jkreft-usgs: Add 'Currently only a "A WMS OGC Web Service" is supported'

jkreft-usgs: Also URL: https://<base_url>/ogcservices/wms

mbucknell: Where do you want that? Should that be in the parameter/value table? And should it be specific, i.e. https://www.waterqualitydata.us/ogcservices/ows ?

PR #236, opened 2017-02-09
Trying to eliminate the tests files from coveralls reports

mbucknell:


mbucknell:
So to omit any python files from coverage consideration you can specify them in the .coveragerc. I have omitted the test files.

PR #238, opened 2017-02-27
use constructor instead of parsing SEARCHPARAMS

jkreft-usgs:
fixes the bug documented in WQP-1076 but does it with copy-pasting.  I don't expect that this pull request will be merged


mbucknell:
Closing. Pull request 239 has a fix for this issue.

PR #240, opened 2017-03-02
Adding required defaults to README

cschroed-usgs:


README.md
```diff
@@ -38,6 +38,19 @@ SEARCH_QUERY_ENDPOINT = ''
 #points to the public srsnames endpoint you want to use.
 PUBLIC_SRSNAMES_ENDPOINT = ''
 
+#points to the geo search endpoint you want to use
+GEO_SEARCH_API_ENDPOINT = ''
```
mbucknell: This is not required. It is in config.py but config.py needs to updated since they have changed the endpoint (again ... without notice)

README.md
```diff
@@ -38,6 +38,19 @@ SEARCH_QUERY_ENDPOINT = ''
...
+
+#a list of dictionaries that associate user-facing text with programmatic style identifiers
+#usually something like:
+# SITE_SLDS = [
+#     {'id' : 'wqp_sources', 'text' : 'By source'},
```
cschroed-usgs: It looks like this README is trying to define as many required config variables as possible, and avoid defining most values. I'm on board with that because it will prevent the README from going stale, and encourage new devs to get to know the configuration repo.

Since the `SITE_SLDS` var is a bit more complex, do you think it makes sense to break this rule and provide an example in the comments?
Is an empty list an appropriate default value, or would an empty list of empty dictionaries be better?

README.md
```diff
@@ -38,6 +38,19 @@ SEARCH_QUERY_ENDPOINT = ''
...
+#     {'id' : 'site_type', 'text' : 'By site type'},
+#     {'id' : 'activity_visual', 'text' : 'By activity'}
+# ]
+SITE_SLDS = []
```
mbucknell: We should probably put these values in config.py. They can be overridden on a tier if that sld is not available.

README.md
mbucknell: In general, I think for a complicated variable like a dictionary it is appropriate to have a comment indicating what is expected.

README.md
```diff
@@ -38,6 +38,14 @@ SEARCH_QUERY_ENDPOINT = ''
 #points to the public srsnames endpoint you want to use.
 PUBLIC_SRSNAMES_ENDPOINT = ''
```
cschroed-usgs: Thanks for your feedback @mbucknell. Is this more what you had in mind?

mbucknell: You might want to say that the id will be used as the sld name and therefore should be available on the geoserver (something like that).

cschroed-usgs: Roger that. Done!

PR #241, opened 2017-03-02
Fixing WQP-1073 by eliminating shared state between mixins.

cschroed-usgs:


test/js/portal_ui/singleClickEventMixinSpec.js
```diff
@@ -0,0 +1,61 @@
+/* jslint browser: true */
+/*global describe */
+/* global beforeEach */
+/* global afterEach */
+/* global it */
+/* global expect */
+/* global PORTAL */
+/* global $ */
+
```
cschroed-usgs: This is a new spec for the slightly-changed implementation.

wqp/portal_ui/static/js/LeafletMixins/SingleClickEventHandlerMixin.js
```diff
@@ -3,62 +3,70 @@
 /* global L */
 /* global _ */
```
cschroed-usgs: Due to indentation changes, the diff in this file is misleadingly large. Here's the skinny:
 - `L.SingleClickEventMixin` used to be an object
 - now `L.SingleClickEventMixin` is a function that returns the object.
 - removed `"use strict"` from each method, placing it instead once in the parent function.

wqp/portal_ui/static/js/LeafletMixins/SingleClickEventHandlerMixin.js
```diff
@@ -3,62 +3,70 @@
...
 
-L.SingleClickEventMixin = {
+/**
+ * This function returns a stateful mixin.
+ */
+L.SingleClickEventMixin = function(){
```
mbucknell: In Leaflet, functions that return new objects typically start with a lowercase letter. We should probably do the same and then it will match the spec file naming too.

cschroed-usgs: Nice! Thanks! Will do!

cschroed-usgs:
Thanks! Tests ftw!

jkreft-usgs:
hurrah!

PR #243, opened 2017-03-14
Wqp 1047 - Refactor the public_srsnames page and eliminate the use of angularjs

mbucknell:
Decided to render the table client side. For now we are just using bootstrap styling and eliminating the ability to sort by column or to narrow the list shown using a search feature. These can be added in the future if needed.


wqp/portal_ui/views.py
```diff
@@ -158,7 +158,10 @@ def other_portal_links():
 def public_srsnames():
     if request.path == '/public_srsnames.jsp':
         return redirect(url_for('portal_ui.public_srsnames-canonical')), 301
-    return render_template('public_srsnames.html')
+
+    resp = session.get(app.config['PUBLIC_SRSNAMES_ENDPOINT'] + '?mimeType=json')
```
ayan-usgs: Not a big deal, but consider using the `param` argument in `requests.get()` in-lieu of concatenating the string. I think it's a bit cleaner that way, but it's your call.

PR #244, opened 2017-03-23
Wqp 1080 - Refactored code that involved Redis caching

mbucknell:
I refactored some of the logic to fetch provider, organization, and site information int wqp/utils.py as well as added a utility to process a tsv from a generator. Added tests for both the wqp/utils.py and wqp/tasks.py. I did not add tests for the views that I updated. That I believe is addressed in another ticket and is outside the scope of WQP-1080 or WQP-1081


wqp/utils.py
```diff
@@ -61,103 +61,168 @@ def geoserver_proxy_request(target_url, cert_verification):
...
-    r = session.get(provider_endpoint, params={"mimeType": "json"})
-    status_code = r.status_code
-    provider_list = None
-    if status_code == 200:
-        codes = r.json().get('codes')
-        provider_list = sorted([code['value'] for code in codes])
-    return {"status_code": status_code, "providers": provider_list}
+    provider_lookups = retrieve_lookups('/providers')
+    if provider_lookups:
+        providers = [code['value'] for code in provider_lookups.get('codes')]
```
mbucknell: I guess this is a question as to how defensive to get with the programming. Lookups should contain the codes keyword even if no codes are returned (it will be the empty array). But I think you are right in that the utility function should return None in that case.

mbucknell:
Improved the utils.py to be able to handle bad response data. Thanks @ayan-usgs for the push!

PR #249, opened 2017-04-17
Wqp 1040

ayan-usgs:


wqp/__init__.py
```diff
@@ -6,9 +9,32 @@
...
+
+    :param str log_directory: optional path of a directory where logs can be written to
+    :return: a logger
+    :rtype: logging.Logger
+
+    """
+    if log_directory is not None:
+        log_file = '{}.log'.format(__name__)
+        log_path = os.path.join(log_directory, log_file)
+        handler = logging.handlers.TimedRotatingFileHandler(log_path, when='midnight', backupCount=20)
```
mbucknell: Consider making the backupCount=30 (one month). We may also want to consider in a separate ticket whether WSI should be managing the cleanup of log files (which I think is what happens with tomcat log files).

wqp/portal_ui/views.py
```diff
@@ -82,6 +82,7 @@ def webservices_documentation():
 @portal_ui.route('/faqs.jsp')
 @portal_ui.route('/faqs/', endpoint='faqs-canonical')
 def faqs():
+    app.logger.info('In the FAQ page.')
```
mbucknell: Is there a reason to log here? It doesn't look like this is a pattern in the other views.

ayan-usgs: No. That was a vanilla trial of logging. I missed it on clean-up. Thanks for catching it!  :octopus:

wqp/__init__.py
```diff
@@ -17,6 +43,45 @@
...
+
+@app.after_request
+def log_after(response):
+    resp_status = response.status
+    resp_headers = response.headers
+    app.logger.debug('Response: {0}, {1}'.format(resp_status, resp_headers))
+    return response
+
+
+def create_request_resp_log_msg(response):
```
mbucknell: Consider moving this and create_redis_log_msg to wqp/utils.py. I'd like to see __init__.py only include objects and functions to set up Flask.

wqp/__init__.py
```diff
@@ -6,9 +9,32 @@
 from flask_swagger_ui import get_swaggerui_blueprint
 from requests import Session
 
+
 __version__ = '4.13.0dev'
 
 
+def create_log_handler(log_directory=None):
```
mbucknell: By defining function here, it makes it available to anything that imports wqp and I don't think we want that. Consider either not making a separate function and just calling what's needed below or prefix it with '_'.

wqp/__init__.py
ayan-usgs: Up-ed to 30.

wqp/__init__.py
ayan-usgs: Moved.

wqp/__init__.py
ayan-usgs: Opted to prefix with `_`.

PR #250, opened 2017-04-17
Wqp 1109 - Develop approach for different WQP UI Skins

mbucknell:
I consider other approaches including using class based views and blueprints. In the end, this seemed to be a more elegant solution as we really just want a different skin and potentially a different set of views for a theme.

The idea is that UI_THEME would be set in the configuration. If 'wqp', the current pages would be served. If 'usgs', then usgs themed pages would be served. I created a decorator to prevent some views from being rendered if 'usgs' is the theme. 

The parent template is now called theme_base.html and is based on the previous base.html with the header, navigation, and footer removed along with the include for the styling. The two theme specific templates extend theme_base.html. The file base.html includes the appropriate theme template based on config.UI_THEME.


dev_install.sh
```diff
@@ -41,6 +41,6 @@ echo "Running Javascript tests";
 node_modules/karma/bin/karma start test/js/karma.conf.js;
 
 echo "Running Python tests";
-env/bin/nosetests
+env/bin/nosetests --logging-clear-handlers
```
ayan-usgs: I'm not seeing a difference with respect to test output when use the `--logging-clear-handlers` option. :thinking:

mbucknell: Is it possible that your instance/config.py is set up with a logging_directory? Your test logging would be going to a file then.

ayan-usgs: `LOGGING_DIRECTORY` is set to `None`.

PR #255, opened 2017-04-18
Reverted version of Flask-Collect because the manage.py collect comma…

mbucknell:
…nd was failing.


ayan-usgs:
Could it be related to this: https://github.com/klen/Flask-Collect/issues/22 ?

mbucknell:
I tried the fix before reverting and it didn't work for me. Rather than continue to mess with it, I reverted the version.

PR #258, opened 2017-04-26
Wqp 1112

ayan-usgs:


wqp/__init__.py
```diff
@@ -54,6 +76,8 @@ def _create_log_handler(log_directory=None):
     # Instead, set the level in the logger object.
     app.logger.setLevel(loglevel)
     app.logger.addHandler(handler)
+    after_setup_logger.connect(_custom_celery_handler)
```
mbucknell: I'm wondering if it's necessary to create two handlers for celery. Does Celery use a different logger for the tasks? I'm wondering if you can use the setup_logging signal and only create the handler once if the logger is the same for the tasks as it is for the rest of celery.

wqp/portal_ui/views.py
```diff
@@ -343,7 +343,7 @@ def sitescachetask(provider_id):
     task = load_sites_into_cache_async.apply_async(args=[provider_id])
     response_content = {'Location': '/'.join([app.config['LOCAL_BASE_URL'], "status", task.id])}
     # passing the content after the response code sets a custom header, which the task status javascript needs
-    return '', 202, response_content
+    return jsonify(response_content), 202
```
mbucknell: This needs to be changed back. The javascript expects the response content in the header. It was out of scope for me to address this as part of the celery refactor.

ayan-usgs: I would lean towards this `jsonfy(response_content), 202, response_content`. The previous code would return an empty response.

wqp/__init__.py
ayan-usgs: I believe that celery does use separate global and task loggers. I can double check, but I'm 97% sure that is the case.

ayan-usgs: Yep. Without the specifying the celery handler on with the `after_setup_task_logger` we miss out on the log statements that from the tasks.

ayan-usgs: I did stumble across some threads where people lamented celery's not quite sane logging implementation. :ghost:

wqp/portal_ui/views.py
mbucknell: That's fine. We should be able to use that when/if we refactor the javascript for the manage_cache page.

ayan-usgs:
@mbucknell:

The response content is back in the headers per your request. Not sure if you want to review again.

wqp/__init__.py
mbucknell: Let's add a comment explaining that celery uses two different loggers. It was not clear from the celery docs in the half hour I spent looking at them.

ayan-usgs: Comment added.

PR #259, opened 2017-04-27
Wqp 1102 - add narrowResult as a result output option

mbucknell:
Also removed the dataProfile query parameter from the  Show Web Service Calls boxes except for the Results box.

### DIFF
--- a/wqp/portal_ui/static/js/views/dataDetailsView.js
+++ b/wqp/portal_ui/static/js/views/dataDetailsView.js
@@ -27,8 +27,8 @@ PORTAL.VIEWS.dataDetailsView = function(options) {
 		var $kml = options.$container.find('#kml');
 
 		var $site = options.$container.find('#sites');
-		var $samples = options.$container.find('#samples');
 		var $biosamples = options.$container.find('#biosamples');
+		var $narrowResults = options.$container.find('#narrowsamples');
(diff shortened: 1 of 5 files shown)


wqp/portal_ui/static/js/views/dataDetailsView.js
```diff
@@ -52,14 +51,14 @@ PORTAL.VIEWS.dataDetailsView = function(options) {
...
+			// If biological results or narrow results desired add a hidden input, otherwise remove it.
+			$dataProfile.remove();
 			if ($biosamples.prop('checked')) {
-				if ($dataProfile.length === 0) {
-					options.$container.append('<input type="hidden" name="dataProfile" value="biological" />');
-				}
+				options.$container.append('<input type="hidden" name="dataProfile" value="biological" />');
+
 			}
-			else {
```
ayan-usgs: Would it be useful to have this follow the `else if` statement that starts on L60?

mbucknell: No ... because it's now removed prior to entering the if-else if clause (see line 55)

ayan-usgs: Ah... missed that.

PR #273, opened 2017-11-27
It is recommended to commit the package-lock.json. This file gets cre…

mbucknell:
…ated automatically when using npm >5.x.x. If npm > 5.x.x is used on the deployment machine npm will generate identical trees, regardless of intermediate dependency updates.

Developers should ensure that they have npm version 5 or greater installed on there development machines and then commit package-lock.json after updates to package.json.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 Water Quality Portal User Interface
 
-This application should be built using python 2.7.x. You also should have node installed on your development machine.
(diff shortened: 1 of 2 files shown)


mbucknell:
The package-lock.json will become more important if we start using CommonJS or ES6 dependencies. Right now it's not technically required, but npm does suggest commiting it to your repo.

PR #274, opened 2017-11-28
Wqp 1167 allow huc 12

ayan-usgs:


jkreft-usgs:
I'm really sorry that i didn't make this explicit in the ticket.  A HUC10 should also be allowed, to follow the existing pattern (2,4,6,8 etc)

ayan-usgs:
> I'm really sorry that i didn't make this explicit in the ticket. A HUC10 should also be allowed, to follow the existing pattern (2,4,6,8 etc)

Okay.

ayan-usgs:
Made some modifications to support HUC10.

ayan-usgs:
Turns out I forgot to push my HUC 10 changes. :gun:

PR #278, opened 2018-01-02
WQP-1170 Select2 for SiteID

skaymen:
https://internal.cida.usgs.gov/jira/browse/WQP-1170

Functioning correctly and passing tests.


mbucknell:
The lookup itself does return sites for WIDNR_WQX. Give that a try.

wqp/portal_ui/static/js/views/siteParameterInputView.js
```diff
@@ -46,6 +46,28 @@ PORTAL.VIEWS.siteParameterInputView = function(options) {
 		});
 	};
 
+	var initializeSiteIdSelect = function($select, organizations) {
+		var formatData = function(data) {
+			if (organizations.length !== 0) { //if an organization is selected, return only those sites
```
mbucknell: This won't work because you need to change the url that is used to fetch the lookup to include organizationid as a query parameter (see the swagger docs). I suspect what needs to happen is that when createPagedCodeSelect is called, it will need an additional option for the current organization id selection and then that will be used to form the url that is used when fetching the look-up.

wqp/portal_ui/static/js/views/siteParameterInputView.js
```diff
@@ -46,6 +46,28 @@ PORTAL.VIEWS.siteParameterInputView = function(options) {
...
+				}
+			}
+			else {
+                return data.value + ' - ' + data.desc;
+            }
+		};
+
+
+		PORTAL.VIEWS.createPagedCodeSelect($select, {
+            codes: 'monitoringlocation',
```
mbucknell: Make sure you are formatting using tabs

wqp/portal_ui/views.py
```diff
@@ -245,7 +245,7 @@ def uri_organization(provider_id, organization_id):
         elif not organization:
             abort(404)
 
-        sites = retrieve_sites_geojson(provider_id, organization_id)
+        sites = retrieve_sites_geojson(provider_id, organization_id);
```
ayan-usgs: Why is there a semi-colon here?

test/js/portal_ui/views/siteParameterInputViewSpec.js
```diff
@@ -40,6 +40,7 @@ describe('Tests for PORTAL.VIEWS.siteParameterInputView', function() {
 			fetch : jasmine.createSpy('organizationFetch').and.returnValue(fetchOrgDeferred)
```
mbucknell: In the beforeEach at line 20, you should be using a <select> tag rather than <input> for <siteid> as that is what it is now.

wqp/portal_ui/templates/portal.html
```diff
@@ -170,7 +170,8 @@
 						<div class="form-group">
 							<label class="control-label" for="siteid">Site ID:</label>
 							<div>
-								<input class="form-control" type="text" name="siteid" id="siteid"/>
+                                <select multiple class="form-control" id="siteid" name="siteid"
```
mbucknell: We use tabs for spacing in the html for this project. Please fix.

test/js/portal_ui/views/siteParameterInputViewSpec.js
```diff
@@ -40,6 +40,7 @@ describe('Tests for PORTAL.VIEWS.siteParameterInputView', function() {
...
 		};
 
+		spyOn(PORTAL.VIEWS, 'createPagedCodeSelect');
```
mbucknell: In the test that starts on line 66, you should add a check to make sure that createPagedCodeSelect was called when the expected arguments.

wqp/portal_ui/static/js/views/portalViews.js
```diff
@@ -31,8 +31,10 @@ PORTAL.VIEWS.createStaticSelect2 = function (el, ids, select2Options) {
  *    @prop {Number} pagesize (optional) - page size to use in request. Defaults to 20
  *    @prop {Function} formatData (optional) - Function takes an Object with value, desc (optional), and providers properties and returns a string.
  * @param {Object} select2Options
+ * @param {Function} func (optional) function used to modify the PagedCodeSelect
```
mbucknell: Consider just passing in the $select jquery element. You can get the value within createPagedCodeSelect by using $select.val(). Make it clear in the comment that you are passing in the jquery element. By convention we usually prepend '$' to the name (so something like $select).

wqp/portal_ui/static/js/views/portalViews.js
```diff
@@ -31,8 +31,10 @@ PORTAL.VIEWS.createStaticSelect2 = function (el, ids, select2Options) {
...
+ * @param {Object} sel (optional) select used in modifying the PagedCodeSelect
  */
-PORTAL.VIEWS.createPagedCodeSelect = function (el, spec, select2Options) {
+PORTAL.VIEWS.createPagedCodeSelect = function (el, spec, select2Options, func, sel) {
```
mbucknell: You should also be passing in the parameter name to use on the lookup (in our case 'organizationid').

wqp/portal_ui/static/js/views/portalViews.js
```diff
@@ -44,6 +46,24 @@ PORTAL.VIEWS.createPagedCodeSelect = function (el, spec, select2Options) {
 		};
 	}
 
+    function geturl(org) {
```
mbucknell: You are assuming that this will always be org. Consider using a more generic name (maybe parentValues). I would also not call this geturl since it doesn't return a url. Maybe getParentParams?

wqp/portal_ui/static/js/views/portalViews.js
```diff
@@ -44,6 +46,24 @@ PORTAL.VIEWS.createPagedCodeSelect = function (el, spec, select2Options) {
...
+    	var orgurl = "";
+    	if (org.length > 0) {
+    		orgurl = "?organizationid=";
+    		orgurl += org.join("&organizationid=");
+    		}
+		return orgurl;
```
mbucknell: Whitespace is off here. Please fix. I think you are using spaces here. Please use tabs for indenting.

wqp/portal_ui/static/js/views/siteParameterInputView.js
```diff
@@ -63,18 +77,32 @@ PORTAL.VIEWS.siteParameterInputView = function(options) {
...
 			initializeOrganizationSelect($organizationSelect, options.organizationModel);
 		});
 
-		// Add event handlers
-		PORTAL.VIEWS.inputValidation({
-			inputEl: $siteIdInput,
-			validationFnc: PORTAL.validators.siteIdValidator
+
+		$organizationSelect.on('change', function(ev) {
+			var organizations = $organizationSelect.val();
```
mbucknell: I think you could put this in createPagedCodeSelect. You already have an event handler in there for the 'change' event.

wqp/portal_ui/templates/portal.html
```diff
@@ -170,7 +170,8 @@
...
+                                        data-placeholder="All" style="..." style="..."></select>
```
mbucknell: I think the two style attributes can both be removed.

wqp/portal_ui/templates/portal.html
skaymen: Done.

wqp/portal_ui/static/js/views/portalViews.js
skaymen: Done.

wqp/portal_ui/templates/portal.html
skaymen: Done.

wqp/portal_ui/static/js/views/portalViews.js
skaymen: Done.

wqp/portal_ui/static/js/views/portalViews.js
skaymen: Done.

wqp/portal_ui/static/js/views/siteParameterInputView.js
skaymen: Done.

wqp/portal_ui/static/js/views/portalViews.js
skaymen: Done.

test/js/portal_ui/views/siteParameterInputViewSpec.js
skaymen: Fixed.

test/js/portal_ui/views/portalViewsSpec.js
```diff
@@ -70,22 +74,28 @@ describe('Tests for PORTAL.VIEWS functions and objects', function () {
...
-			PORTAL.VIEWS.createPagedCodeSelect($('#test-div'), testSpec, {placeholder: 'Pick one'});
+			PORTAL.VIEWS.createPagedCodeSelect($('#test-select2'), testSpec, {placeholder: 'Pick one'});
 			var options = $.fn.select2.calls.argsFor(0)[0];
 			expect(options.allowClear).toEqual(true);
 			expect(options.placeholder).toEqual('Pick one');
+		});
 
+		it('Expects the filter select parameters will be used when provided', function () {
+			PORTAL.VIEWS.createPagedCodeSelect($('#test-select2'), testSpec, {}, testFilter, testParam);
+			testFilter.val(testParam).trigger('change');
```
mbucknell: You shouldn't use the same text for the parameter name and the value of the input. How about setting this to "param1" and then test below for test=param1

wqp/portal_ui/static/js/views/portalViews.js
```diff
@@ -31,11 +31,12 @@ PORTAL.VIEWS.createStaticSelect2 = function (el, ids, select2Options) {
  *    @prop {Number} pagesize (optional) - page size to use in request. Defaults to 20
  *    @prop {Function} formatData (optional) - Function takes an Object with value, desc (optional), and providers properties and returns a string.
  * @param {Object} select2Options
+ * @param {jquery element} $sel (optional) select used in filtering the PagedCodeSelect
```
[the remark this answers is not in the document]
skaymen: Done.

wqp/portal_ui/static/js/views/portalViews.js
```diff
@@ -44,6 +45,31 @@ PORTAL.VIEWS.createPagedCodeSelect = function (el, spec, select2Options) {
 		};
 	}
 
+	function getParentParams(parentValue) {
+		var suffix = "";
+		//add parentValue to URL, using .join if it is an array and simply appending if a string
+		if (parentValue.length > 0) {
+			suffix = "?" + parametername + "=";
+			if (typeof parentValue !== "string") {
```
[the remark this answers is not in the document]
skaymen: Done.

wqp/portal_ui/static/js/views/portalViews.js
```diff
@@ -31,10 +31,10 @@ PORTAL.VIEWS.createStaticSelect2 = function (el, ids, select2Options) {
  *    @prop {Number} pagesize (optional) - page size to use in request. Defaults to 20
  *    @prop {Function} formatData (optional) - Function takes an Object with value, desc (optional), and providers properties and returns a string.
  * @param {Object} select2Options
- * @param {jquery element} $sel (optional) select used in filtering the PagedCodeSelect
+ * @param {jquery element} $sel (optional) input, select, or textarea used in filtering the PagedCodeSelect
```
mbucknell: Typo $sel is now $filter

PR #289, opened 2018-06-07
Wqp 1158 - Integrate internal WQP UI with Water-Auth

mbucknell:
The code assumes that the authentication service uses Oauth2. Whether its used or not is determine by configuration variables (WATERAUTH_*). On the client side, the presence of the access_token determines whether the Authorization header is set in Ajax calls and whether the download proxy is used rather than direct connection to the download service.

Development was done using a local authentication service and running all WQP services locally. This required that the development runs with https. See the server/README.md for instructions on how to do this.


assets/js/leafletLayers/WQPSitesLayer.js
```diff
@@ -43,6 +43,37 @@
 			return this.queryParamArray;
 		},
 
+		_getImageSrc : function(url, done) {
```
mbucknell: Now the layer image is fetched as a blob so that we can set the authorization header. The bolob as then read as url data and assigned to the src attribute of the layer img tag.

server/wqp/portal_ui/templates/theme_base.html
```diff
@@ -42,6 +42,20 @@
 				ResultDetectionQuantitationLimit: Config.SEARCH_QUERY_ENDPOINT + 'ResultDetectionQuantitationLimit/search',
 				biologicalresult: Config.SEARCH_QUERY_ENDPOINT + 'biologicalresult/search'
 			};
+			if ("{{ config.WATERAUTH_AUTHORIZE_URL }}") {
```
mbucknell: We only want to use the proxy if authorization is required.

server/wqp/portal_ui/views.py
```diff
@@ -169,6 +178,21 @@ def public_srsnames():
...
+    Proxies the download request and adds the authorization header if an access_token is present.
+    :param String op: The kind of download to request
+    :return Response:
+    '''
+    target_url = app.config['SEARCH_QUERY_ENDPOINT'] + op + '/search'
+    headers = {}
+    access_token = request.cookies.get('access_token')
+    if access_token:
+        headers['Authorization'] =  'Bearer {0}'.format(access_token)
+    resp = session.post(target_url, data=request.form, headers=headers, verify=proxy_cert_verification)
```
ayan-usgs: I'm not seeing where `proxy_cert_verification` is defined....

server/wqp/portal_ui/views.py
```diff
@@ -169,6 +178,21 @@ def public_srsnames():
     return render_template('public_srsnames.html', status_code=resp.status_code, content = resp.json())
 
 
+@portal_ui.route('/wqp_download/<op>', methods=['GET', 'POST'])
+def wqp_download_proxy(op):
```
ayan-usgs: Does this have a test?

server/wqp/auth/views.py
```diff
@@ -0,0 +1,54 @@
...
+    :return: Decorated function
+    :rtype: function
+    '''
+    @wraps(f)
+    def decorated_function(*args, **kwargs):
+        '''
+        Function which redirects to the authorization endpoint if the access token has expired otherwise
+        executes the function.
+        '''
+        if app.config['WATERAUTH_AUTHORIZE_URL'] and (session.get('access_token_expires_at', 0) < int(time.time())):
```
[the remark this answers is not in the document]
mbucknell: The access_token_expires_at in the session is an integer.

server/wqp/portal_ui/views.py
mbucknell: Earlier on line 31. It's assigned ```proxy_cert_verification = app.config.get('PROXY_CERT_VERIFY', False)```

server/wqp/portal_ui/views.py
mbucknell: It does not. I'll add some. I was having issues with the python tests this morning and meant to come back to this but did not.

server/wqp/portal_ui/views.py
ayan-usgs: Ah, yes. It was collapsed.

server/wqp/auth/views.py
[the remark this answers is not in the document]
ayan-usgs: I guess that no reason to sweat the loss of some microseconds in the precision. :stuck_out_tongue:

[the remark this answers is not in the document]
mbucknell: Also the JWT standard is fuzzy and says that it is okay to accept an expired token if it's only a few seconds expired, so there is already fuzziness.

mbucknell:
I'm pretty sure I still have whitespace issues but I'm going to merge and sort out later.

PR #296, opened 2018-06-08
Trying to add python to coveralls

mbucknell:


mbucknell:
This now brings in the python code to coveralls. It's curious that the percent covered is so much higher than before we changed the project structure.

PR #299, opened 2018-06-11
Fix bug with Show Web Service Calls and Show ArcGIS Online Parameters

mbucknell:


mbucknell:
I don't think there will be because things are fairly self contained. But the showAPIView is one of those spots where it's not.

PR #301, opened 2018-06-13
Wqp 1217 - Linkable filled out portal page

mbucknell:
The idea is that we can initialize the form via query parameters after the # in the url. Added initValues to all of the 4 flavors of select2s in portalViews. For text inputs, added a utility function to initialize the value of an input from the hash. Hidden inputs had to be handled individually (bbox, sorted, and nldiurl) and a 'change' event is triggered whenever its value is updated so that the url is updated with the new query parameters. The radio buttons for File format also had to be handled individually.


assets/test/js/portal_ui/utilsSpec.js
```diff
@@ -97,4 +98,58 @@ describe('Test PORTAl.UTILS package', function () {
...
+
+    describe('getAnchorQueryValue', () =>  {
+        it('Return the empty array if there is no anchor part of the url', () => {
+            window.location.hash = '';
+
+            expect(getAnchorQueryValues('name1')).toEqual([]);
+        });
+
+        it('Return the empty array if the anchor part does not contain the parameter name', () => {
+           window.location.hash = '#name2=val1&name3=val2';
```
ayan-usgs: Is it necessary to have a case where a key doesn't have a value if someone manually edits the URL? For example, `#name1=val1&name2=&name3=val3`.

mbucknell:
I'm going to go back and address Codacy issues tomorrow. Wanted to stop delaying working on RDSR 😁

assets/test/js/portal_ui/utilsSpec.js
mbucknell: It would be good to add a test for that.

PR #304, opened 2018-06-18
Wqp 1225 - Shareable url box

mbucknell:


server/wqp/portal_ui_blueprint/templates/portal.html
```diff
@@ -466,6 +466,10 @@
                         <input type="hidden" name="sorted" id="hidden-sorted" value="no"/>
                         <div id="form-action">
                             <button id="main-button" class="btn" type="submit">DOWNLOAD</button>
+                            <div class="share-container">
+                                <button type="button" class="btn">Share</button>
```
danielnaab: I wonder if "Copy to Clipboard" rather than "Share" would be more clear.

mbucknell: agreed

PR #307, opened 2018-06-20
Wqp 1195 show only one webcall

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WQP-1195 Show Only One Web Service Call for the Portal Page
-----------

For https://internal.cida.usgs.gov/jira/browse/WQP-1195 , the response of the 'Show Web Calls' button was changed from showing many calls to showing a single call based on which 'Select data to download' radio button was selected by the user. 

In addition a text box was added to show a 'curl' command (when the 'Show Web Calls' button is pressed) based on the user selected form information.

-----------

After making a pull request
---------------------------


CHANGELOG.md
```diff
@@ -8,5 +8,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Copy to clipboard button that takes the current portal page URL that is the window below the button and puts it in
 the browser clipboard.
+- Curl command that shows when the 'Show Web Calls' button is clicked
+
+### Changed
+- Function of 'Show Web Calls' button so that only one call, based on which radio button is selected, is shown
```
mbucknell: Could you say "result type radio button" is selected?

assets/js/portalOnReady.js
```diff
@@ -15,33 +14,34 @@ $(document).ready(function () {
...
+    let arcGisOnlineHelpView = new ArcGisOnlineHelpView({
         $button : $('#show-arcgis-online-help'),
         $dialog : $('#arcgis-online-dialog'),
         $siteMapViewContainer : $('#mapping-div'),
         getQueryParamArray : $.proxy(downloadFormView.getQueryParamArray, downloadFormView)
     });
 
     //Initialize subviews
-    var initDownloadForm = downloadFormView.initialize();
+    let initDownloadForm = downloadFormView.initialize($.proxy(showAPIView.updateWebCallDisplay, showAPIView), $.proxy(showAPIView.hideDivOnChange, showAPIView)); // new line for WQP-1195
```
mbucknell: I don't think the second parameter is used in the initialize call (see https://github.com/abriggs-usgs/WQP_UI/blob/86b98d03c4c50c685a35f5c737917d0359bf4368/assets/js/views/downloadFormView.js#L72). And please remove the trailing comment.

assets/js/utils.js
```diff
@@ -168,3 +168,77 @@ export const initializeInput = function($el) {
     const initValues = getAnchorQueryValues($el.attr('name'));
     $el.val(initValues.length ? initValues[0] : '');
 };
+
+/*
+ *  Creates a constant with properties indicating if a data profile is used
+ *  @return {object} a data profile object with boolean properties indicating whether data profile is used
+ */
+export const checkForUseOfDataProfileArray = function () {
```
mbucknell: Consider just exporting  a const object. This doesn't really need to be a function call.

assets/js/utils.js
```diff
@@ -168,3 +168,77 @@ export const initializeInput = function($el) {
...
+ * @return {string} The formatted string used to make curl calls.
+ */
+export const separateCurlDataFromParams = function (queryParamArray) {
+    let allParams = {};
+    allParams.curlParamsArray = [];
+    allParams.curlDataArray = [];
+
+    // separate data parameters
+    each(queryParamArray, function (param) {
+        if (param['name'] === 'mimeType' || param['name'] === 'zip' || param['name'] === 'sorted') {
```
mbucknell: Consider creating a constant array containing the "data" parameters. Then you could use lodash includes (https://lodash.com/docs/4.17.10#includes) to write as ```if (includes(DATA_PARAMS, param.name))```. Makes it easier to add and remove data parameters as needed to.

assets/js/utils.js
```diff
@@ -168,3 +168,77 @@ export const initializeInput = function($el) {
...
+            allParams.curlDataArray.push(param);
+        }
+    });
+    return allParams;
+};
+
+/*
+ * Assembles a curl string from the user entered form values
+ * @param resultType {string} The result returned when the user selects form elements.
+ * @return {string} a formatted line that can be used a curl command.
```
mbucknell: You need to document the second parameter. Particularly since the function expects it to be an object with specific parameters.

assets/js/utils.js
```diff
@@ -168,3 +168,77 @@ export const initializeInput = function($el) {
...
+    let curlLeadingString = 'curl -X POST --header \'Content-Type: application/json\' --header \'Accept: application/';
+    let urlBase = Config.QUERY_URLS[resultType];
+    let params = $.param(allParams.curlParamsArray);
+
+    let dataParamsString = '';
+    let mimeTypeValue = '';
+
+    // create a string of formatted parameters
+    each(allParams.curlDataArray, function(param) {
+        let partialDataParamsString = `"${param.name}":["${param.value}"]`;
```
mbucknell: Does this work when you speciify minresults or nldiurl? These parameters should be strings in the post json object. See https://cida-eros-wqpdev.er.usgs.gov:8443/qw_portal_services/swagger-ui.html#!/Station/stationJsonPostRequestUsingPOST.  

Take a look at getQueryParamJson and how it's used here: https://github.com/NWQMC/WQP_UI/blob/f2e522027da49d0c509aec320f664153ca9704be/assets/js/queryService.js#L25. I think you can eliminate separateCurlDataFromParams by using a similar pattern.

assets/js/views/downloadFormView.js
```diff
@@ -69,7 +69,7 @@ export default class DownloadFormView {
      *      @resolve - if all initialization including successful fetches are complete
      *      @reject - if any fetches failed.
      */
-    initialize() {
+     initialize(updateWebCallDisplay) {
```
mbucknell: Please update the comment to include this parameter. Also consider a more generic name for the parameter, such as updateResultTypeAction

assets/js/views/showAPIView.js
```diff
@@ -1,44 +1,56 @@
...
+            let resultType = this.getResultType();
+            let queryParamArray = this.getQueryParamArray();
+            let queryWithoutDataProfileArray = queryParamArray.filter((param) => {
+               return param.name !== 'dataProfile';
             });
-            var queryString = getQueryString(queryParamArray);
-            var queryStringWithoutDataProfile = getQueryString(queryWithoutDataProfileArray);
+
+            let queryString = getQueryString(queryParamArray);
+            let queryStringWithoutDataProfile = getQueryString(queryWithoutDataProfileArray);
```
mbucknell: It's a little unclear to me why dataProfile has to be removed. It should only be set to a value when required.

assets/js/views/showAPIView.js
```diff
@@ -1,44 +1,56 @@
...
-            $activitymetricsText.html(queryService.getFormUrl('ActivityMetric', queryStringWithoutDataProfile));
-            $resultdetectionText.html(queryService.getFormUrl('ResultDetectionQuantitationLimit', queryStringWithoutDataProfile));
+            $apiQueryTitle.html(resultType.replace(/([A-Z])/g, ' $1'));
+            $apiQueryText.html(apiQueryString);
+            $cUrlText.html(curlString);
             $wfsText.html(L.WQPSitesLayer.getWfsGetFeatureUrl(queryWithoutDataProfileArray));
         });
     }
+
+    updateWebCallDisplay() {
```
mbucknell: This doesn't do anything. Why is it here?

CHANGELOG.md
abriggs-usgs: Sure can.

assets/js/portalOnReady.js
abriggs-usgs: Ah, the dangers of long lines--didn't see that.

assets/js/views/showAPIView.js
abriggs-usgs: Hmm. Yeah, I was trying to touch as little of the old code as possible. But I see now it was only written that way to show both the calls with and the calls without at the same time, which is no longer needed. I'll fix it.

assets/js/utils.js
abriggs-usgs: Correct. Good advice.

assets/js/utils.js
abriggs-usgs: Good call again. Done.

assets/js/utils.js
abriggs-usgs: Fixed

assets/js/utils.js
abriggs-usgs: Fixed

assets/js/views/downloadFormView.js
abriggs-usgs: Fixed

assets/js/utils.js
```diff
@@ -168,3 +169,40 @@ export const initializeInput = function($el) {
...
+
+
+/*
+ * Assembles a curl string from the user entered form values
+ * @param resultType {string} The value gathered from user input on the web form.
+ * @param queryParamArray An array of values gathered from user input on web form.
+ * @return {string} a formatted line that can be used a curl command.
+ */
+export const buildCurlString = function(resultType, queryParamArray) {
+    let curlLeadingString = 'curl -X POST --header \'Content-Type: application/json\' --header \'Accept: application/';
```
mbucknell: The valid values for Accept header mimeType is not always 'application/whatever'. For instance, the valid mimeType for csv is text/csv.

assets/js/utils.js
```diff
@@ -168,3 +169,40 @@ export const initializeInput = function($el) {
...
+    let urlBase = Config.QUERY_URLS[resultType];
+    let queryParamJson = getQueryParamJson(queryParamArray);
+    let dataParameters = omit(queryParamJson, ['mimeType', 'zip', 'sorted']);
+    let queryParameters = pick(queryParamJson, ['mimeType', 'zip', 'sorted']);
+    let params = $.param(queryParameters);
+    let mimeTypeValue = pick(queryParameters, ['mimeType'])['mimeType'];
```
mbucknell: I think you can more simply use queryParameters.mimeType.

assets/js/views/showAPIView.js
```diff
@@ -1,43 +1,50 @@
...
+        let $apiQueryText = this.$container.find('#query-div textarea');
+        let $cUrlText = this.$container.find('#curl-query-div textarea');
+        let $wfsText = this.$container.find('#getfeature-query-div textarea');
 
         this.$container.find('#show-queries-button').click(() => {
-            var queryParamArray = this.getQueryParamArray();
-            var queryWithoutDataProfileArray = queryParamArray.filter((param) => {
-                return param.name !== 'dataProfile';
+            let resultType = this.getResultType();
+            let queryParamArray = this.getQueryParamArray();
```
mbucknell: So I still don't think you need to do this. If dataProfile is present in the form when it needs to be and is not there if it is not needed. This just unnecessarily complicates the code.

assets/js/utils.js
```diff
@@ -168,3 +169,40 @@ export const initializeInput = function($el) {
...
 };
+
+
+/*
+ * Assembles a curl string from the user entered form values
+ * @param resultType {string} The value gathered from user input on the web form.
+ * @param queryParamArray An array of values gathered from user input on web form.
+ * @return {string} a formatted line that can be used a curl command.
+ */
+export const buildCurlString = function(resultType, queryParamArray) {
```
mbucknell: Can we call this getCurlString? That follows the pattern that we are using with functions like getQueryParams, etc.

assets/test/js/portal_ui/utilsSpec.js
```diff
@@ -159,4 +159,28 @@ describe('Test PORTAl.UTILS package', function () {
...
+
+        let testResultType = 'Station';
+        let testQueryParamArray = {};
+        testQueryParamArray = [
+            {name: 'mimeType', value: 'mimeTypeTestValue'},
+            {name: 'zip', value: 'zipTypeTestValue'},
+            {name: 'sorted', value: 'sortedTestValue'},
+            {name : 'statecode', value : ['US:55', 'US:54'], multiple: false},
+            {name : 'huc', value: '0701*;0702*', multiple: true},
+            {name : 'siteType', value : 'Well', multiple: true}
```
mbucknell: It would be great to include a parameter that is coded as a string, such as startDateLo

assets/js/utils.js
abriggs-usgs: Okay. Did the best that I could with this. The Swagger Docs were inconsistent, and I couldn't find a full list of types anywhere.

mbucknell: Here's an official list https://www.freeformatter.com/mime-types-list.html

assets/js/utils.js
abriggs-usgs: Hmm . . . I tried this, let mimeTypeValue2 = contentType[pick(queryParameters.mimeType)]; , but could not get it to work.

assets/js/utils.js
abriggs-usgs: Yup. Done.

assets/js/utils.js
mbucknell: The other thing is that I just noticied is that if zip is selected the Accept is application/zip. You might want to ask Jim if you should ignore zip when forming the curl. If not this becomes super easy.

assets/js/utils.js
mbucknell: Did you try ``` mimeTypeValue2 = queryParameters.mimeType? Isn't queryParameters a json object.

assets/js/utils.js
mbucknell: I would just set it to application/zip as that matches the GET command above. So that solves that problem!

jkreft-usgs: I would like it to hew as closely to the actual form as possible.  One thing though is that I think you could just drop the accept headers all together, or use */*, since the mimetype is really set as a query param

jkreft-usgs: oh that's right, you can't choose NOT to zip it on the query form, so just go for that

assets/js/utils.js
abriggs-usgs: Oh, yes, I did try that. Sorry, I changed the line of code from what was in the original post (duh), that made things more confusing. Anyway, maybe I miss understand the way Lodash 'pick()' works but I think I may need to have object followed by a comma -- pick(object, [key]). At least I can't get it to work any other way. This is the an example of the object: [{"name":"mimeType","value":"tsv","multiple":false},{"name":"zip","value":"yes","multiple":false}]

mbucknell: When I debug, queryParameters is {'mimeType:'csv', "zip:"yes"}. Maybe you can stop by and we can both look at this.

PR #323, opened 2018-10-01
Wqp 1247 - Identify dialog downloads are incomplete and the dataProfile ones don't work

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

WQP 1247
-----------
Identify Dialog now populates the form on button click rather than have a form within the dialog

Description
-----------
Decided to implement using a Leaflet popup. The dialog has the same content on mobile.

After making a pull request
---------------------------


assets/js/identifyDialog.js
```diff
@@ -1,97 +1,54 @@
...
-            if (showOptions.features.length > 0) {
-                $detailDiv.html(featureInfoTemplate(context));
-                $hiddenFormInputDiv.html(hiddenFormTemplate(context));
 
-                this.$dialog.dialog('open');
-            } else {
-                this.$dialog.dialog('close');
             }
-        }
+        });
```
danielnaab: I like how concise this code is. Would it make sense to clear the existing search terms before adding the new options?

assets/js/hbTemplates/identifyDialog.hbs
```diff
@@ -1,5 +1,8 @@
+<div class="button-container">
+	<button id="identify-populate-button" class="btn btn-default" type="button">Populate form</button>
+</div>
```
danielnaab: It wasn't immediately obvious to me how this button was intended to behave - maybe a different wording could be more explicit. Thoughts - "Filter by selection", "Add selection to filter form"...

mbucknell: I agree the name of the button is not good. I think Filter by selection is too generic. I wonder if something like Add Sites to Form would work even though in some cases we are filling out the bounding box parameter?

assets/js/identifyDialog.js
mbucknell: Strictly speaking, we are adding the identified sites to the form so I don't think clearing would be the correct thing to do. In particular, when we use the bounding box parameter we can not clear the search terms.

PR #329, opened 2018-11-28
update coverage badge, cleaned up python requirements.

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Updated badge, fixed Makefile for server, and cleaned up Python requirements
-----------

After making a pull request
---------------------------


mbucknell:
Agree that we should be looking to manage our python dependencies better, but would like to not do it on WQP-UI until it gets moved to the cloud. Keep reminding me though. Maybe we can write a ticket for NGWMN to explore dependancy management tools.

PR #376, opened 2019-02-12
Updated to latest version of karma-rollup…

mbucknell:
…-preprocessor and eliminated temp-test-manifest.js

Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Using the latest version of karma-rollup-preprocessor allow us to remove temp-test-manifest.js

Description
-----------

After making a pull request
---------------------------


mbucknell:
I updated karma, karma-jasmine, and karma-rollup-preprocessor to the latest. I eliminated the temp-test-manifest.js and tested npm run test:watch to make sure tests were rerun as appropriate. All seems to work.

assets/karma.conf.js
```diff
@@ -26,7 +26,11 @@ module.exports = function (config) {
         files: [
             'dist/scripts/vendor.js',
             'test/resources/testConfig.js',
-            'temp-test-manifest.js',
+            'js/leafletControls/*.js',
+            'js/leafletLayers/*.js',
+            'js/LeafletMixins/*.js',
+            'js/leafletUtils.js',
+            {pattern: 'test/js/**/*.js', watched: false},
```
danielnaab: Does this cause any speed issues you've seen? The karma rollup preprocessor doesn't seem to handle lots of modules very well, compiling each test file into a separate bundle, rather than combining them into one in some manner. Wondering if that might be an issue here?

mbucknell: Initial test start up is a bit slow but I don't think it's a dramatic difference. The watch seems to work correctly in that only the changed bundle get's re bundled. I would rather have a slightly slower start-up vs maintaining a separate temp-test-manifest file. YMMV

mbucknell: This also solves the problem that I have observed where changing the code that's being tested doesn't cause the tests to rerun when using watch. I will do a little more work to see if the issue with leaflet can be solved some other way other than explictly loading those files.

mbucknell:
These changes do increase the time for bundling. On my laptop it was roughly 30 seconds. But if you are using npm run test:watch, it will no rerun tests when changing either specs or code.

PR #523, opened 2019-08-20
Wqp 1516 - Maps aren't loading for some reason

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Renames map div to site-map and sites-map so that the hashing process doesn't change the name of the id we are using in the javascript.

Description
-----------
I also removed the symbolic link static. We will see if this breaks deployment at CHS.

After making a pull request
---------------------------

### DIFF
--- a/assets/package.json
+++ b/assets/package.json
@@ -13,7 +13,6 @@
     "test:watch": "mkdir -p dist/scripts && run-p build:vendorjs && karma start --debug --no-single-run",
     "build": "run-p build:* && npm run hash-assets",
     "build:css": "run-p buildcss:*",
-    "buildcss:site": "lessc less/site.less > dist/site.css && uglifycss --output dist/site.css dist/site.css && postcss dist/site.css --no-map --use autoprefixer -o dist/site.css",
     "buildcss:theme_usgs": "lessc less/usgs_theme/custom.less > dist/nwis_qw_custom.css && uglifycss --output dist/nwis_qw_custom.css dist/nwis_qw_custom.css && postcss dist/nwis_qw_custom.css --no-map --use autoprefixer -o dist/nwis_qw_custom.css",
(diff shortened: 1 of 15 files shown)


assets/package.json
```diff
@@ -13,7 +13,6 @@
     "test:watch": "mkdir -p dist/scripts && run-p build:vendorjs && karma start --debug --no-single-run",
     "build": "run-p build:* && npm run hash-assets",
     "build:css": "run-p buildcss:*",
-    "buildcss:site": "lessc less/site.less > dist/site.css && uglifycss --output dist/site.css dist/site.css && postcss dist/site.css --no-map --use autoprefixer -o dist/site.css",
```
mbucknell: By not building a separate site.css file, this would have fixed the issue without renaming the id, but I decided to rename anyways to follow best practices for id names and use the dash rather than underscore.

server/wqp/portal_ui_blueprint/templates/sites.html
```diff
@@ -2,13 +2,6 @@
 
 {% block title %}Water Quality Portal Data Sites for {{ organization }}{% endblock %}
 
-{% block page_style %}
```
mbucknell: This seemed unnecessary as all other pages only use custom.css. That's why I removed a separate site.css file.

assets/package.json
ayan-usgs: :+1:

server/wqp/sites/utils.py
```diff
@@ -141,7 +141,11 @@ def site_feature_generator(iter_lines):
     found_header = False
     headers = []
     while not found_header:
-        line = next(iter_lines)
+        try:
+            line = next(iter_lines)
+        except StopIteration as e:
+            app.logger.warning('Received an empty rdb response')
```
ayan-usgs: Not a huge deal, but I'm not sure `StopInteration as e` is necessary because exception attributes aren't being accessed.

PR #631, opened 2019-11-26
Update dependencies

mbucknell:
Before making a pull request
----------------------------

- [X] Make sure all tests run

Description
-----------
Note that we are still not updating karma-rollup-preprocessor because it causes most of the jasmine tests to not execute. In addition we are not updating to the latest version of rollup (1.27.x) because of a change in how plugins are integrated which is causing a lot of warnings to be raised. I think we should wait for more of the plugins to be updated to handle this appropriately.

After making a pull request
---------------------------


server/requirements.txt
```diff
@@ -1,14 +1,14 @@
...
 beautifulsoup4==4.8.1
 celery==4.3.0
 feedparser==5.2.1
 Flask==1.1.1
 Flask-WTF==0.14.2
 flask-swagger==0.2.14
 geojson==2.5.0
-gunicorn[gevent]==19.9.0
-pyproj==2.4.0
+gunicorn[gevent]==20.0.3
```
mbucknell: @ayan-usgs, Only change I observed is that gunicorn shows a warning if the config file is not a python module (ends with .py). So I renamed gunicorn.conf to gunicorn.conf.py

PR #894, opened 2021-03-16
Updated node version in Assets readme; Added Windows instructions to …

lprivette:
…readme

 [ x ] Make sure all tests run
 [  ] Update the changelog appropriately (not necessary for this change?)

Updates to READMEs
-----------
- Updated readmes to address windows machines attempting to setup the project and a no longer compatible node version.
- Jira ticket: WQP-1642

Description
-----------
- In 'assets' readme, updated node version. 8.11.2 to 10.18.1 due to error when attempting to run the project. 
- In main readme, added info on installing WSL for windows machines setting up the project.

After making a pull request
---------------------------

### DIFF
--- a/assets/README.md
+++ b/assets/README.md
@@ -5,12 +5,12 @@ produced by this project are dependencies of the [Flask server](../server).
 
 ## Install dependencies
 
-Javascript and LESS assets are built with Node.js v10.18.1. Usage of
+Javascript and LESS assets are built with Node.js. Check the package.json for the specific version. Usage of
 [nvm](https://github.com/creationix/nvm) is a convenient way to use a specific
 version of Node.js:
 
 ```bash
-nvm use v10.18.1
+nvm use v12.18.1
 ```
 
 Node.js dependencies are installed via:
--- a/assets/README.md
+++ b/assets/README.md
@@ -5,12 +5,12 @@ produced by this project are dependencies of the [Flask server](../server).
 
 ## Install dependencies
 
-Javascript and LESS assets are built with Node.js v8.11.2. Usage of
+Javascript and LESS assets are built with Node.js v10.18.1. Usage of
 [nvm](https://github.com/creationix/nvm) is a convenient way to use a specific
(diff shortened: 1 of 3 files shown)


assets/README.md
```diff
@@ -5,12 +5,12 @@ produced by this project are dependencies of the [Flask server](../server).
 
 ## Install dependencies
 
-Javascript and LESS assets are built with Node.js v8.11.2. Usage of
+Javascript and LESS assets are built with Node.js v10.18.1. Usage of
```
mbucknell, line 8: If the make scripts are used the version of node is taken from assets/package.json so it's "engines": {
    "node": "12.18.1"
  }. I would update this to say v12.* or maybe just say explictly look in assets/package.json for the version to use.

lprivette: Mitch is running it using 10.18.1 and that's the version I originally used before re-cloning and setting everything up again to fix that weird issue with the template changes not updating. Does that change your opinion at all? I guess it would be best to just use the later versions if possible.

mbucknell: So you aren't using the make env? That should check to see what version of node that you are using and fail if you do not have the version it wants. But yes, I think it would be best to use the latest LTS version.

lprivette: No we definitely did the `make devenv`. So that's correct for my most recent build and why I'm at 12.18.1. But now I'm wondering how both Mitch and I ended up at 10.18.1 when the `make devenv` would have installed the 12.18.1 listed in the package.json. I bet during our first time setting up the whole environment we tried changing node version to the 8.11.2 specified in the readme after encountering some error. Sorry for all the confusion -- I will update text to refer to the version listed in the package.json

mbucknell: Just to be clear make devenv does not install the correct version of node. I manage my versions of node using nvm

lprivette: Ok, I think I'm following. I've installed nvm too

mbucknell (approved):

PR #961, opened 2021-06-07
Corrects tooltips test; Changes 'STORET' references to 'EPA' (Wqp 1671)

lprivette:
- [x] Make sure all tests run

Changes 'STORET' references to 'EPA' on presentation side NOT services

Corrects providers not updating in url and tests 

-----------
https://internal.cida.usgs.gov/jira/browse/WQP-1671

After making a pull request
---------------------------

### DIFF
--- a/assets/js/hbTemplates/coverageMapPopup.hbs
+++ b/assets/js/hbTemplates/coverageMapPopup.hbs
@@ -9,7 +9,7 @@
         </div>
         {{#if EPA_DISCRETE_SAMPLE_COUNT}}
             <div>
-                <b>EPA STORET discrete samples: </b>{{EPA_DISCRETE_SAMPLE_COUNT}}
+                <b>EPA EPA discrete samples: </b>{{EPA_DISCRETE_SAMPLE_COUNT}}
             </div>
         {{/if}}
         {{#if NWIS_DISCRETE_SAMPLE_COUNT}}
--- a/assets/js/hbTemplates/coverageMapPopup.hbs
+++ b/assets/js/hbTemplates/coverageMapPopup.hbs
@@ -9,7 +9,7 @@
         </div>
         {{#if EPA_DISCRETE_SAMPLE_COUNT}}
             <div>
-                <b>EPA STORET discrete samples: </b>{{EPA_DISCRETE_SAMPLE_COUNT}}
+                <b>EPA discrete samples: </b>{{EPA_DISCRETE_SAMPLE_COUNT}}
             </div>
         {{/if}}
         {{#if NWIS_DISCRETE_SAMPLE_COUNT}}
--- a/assets/js/hbTemplates/coverageMapPopup.hbs
+++ b/assets/js/hbTemplates/coverageMapPopup.hbs
@@ -9,7 +9,7 @@
         </div>
         {{#if EPA_DISCRETE_SAMPLE_COUNT}}
             <div>
-                <b>EPA EPA discrete samples: </b>{{EPA_DISCRETE_SAMPLE_COUNT}}
+                <b>EPA STORET discrete samples: </b>{{EPA_DISCRETE_SAMPLE_COUNT}}
             </div>
         {{/if}}
         {{#if NWIS_DISCRETE_SAMPLE_COUNT}}
--- a/assets/test/js/portal_ui/queryServiceSpec.js
+++ b/assets/test/js/portal_ui/queryServiceSpec.js
@@ -41,7 +41,7 @@ describe('Tests for queryService', function() {
 
         it('Expects that the mimeType, zip, and sorted are removed from the json payload and that a POST request is made', function() {
             var requestBody;
-            queryService.fetchQueryCounts('Station', testQuery, ['NWIS', 'EPA']);
+            queryService.fetchQueryCounts('Station', testQuery, ['NWIS', 'STORET']);
 
             expect(fakeServer.requests.length).toBe(1);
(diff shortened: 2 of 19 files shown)


assets/js/hbTemplates/coverageMapPopup.hbs
```diff
@@ -9,7 +9,7 @@
         </div>
         {{#if EPA_DISCRETE_SAMPLE_COUNT}}
             <div>
-                <b>EPA STORET discrete samples: </b>{{EPA_DISCRETE_SAMPLE_COUNT}}
+                <b>EPA EPA discrete samples: </b>{{EPA_DISCRETE_SAMPLE_COUNT}}
```
mbucknell, line 12: Probably just EPA discrete samples

assets/test/js/portal_ui/queryServiceSpec.js
```diff
@@ -61,11 +61,11 @@ describe('Tests for queryService', function() {
         });
 
         it('Expects a successful response passes a correctly formatted counts record', function() {
-            queryService.fetchQueryCounts('Result', testQuery, ['NWIS', 'STORET']).done(successSpy).fail(errorSpy);
+            queryService.fetchQueryCounts('Result', testQuery, ['NWIS', 'EPA']).done(successSpy).fail(errorSpy);
             fakeServer.respondWith([200, {'Content-Type': 'application/json'},
                 '{"NWIS-Site-Count":"492","Total-Site-Count":"492","NWIS-Result-Count":"6641","Total-Result-Count":"6641",' +
                 '"NWIS-Activity-Count":"664","Total-Activity-Count":"664",' +
-                '"STORET-ActivityMetric-Count": "232", "Total-ActivityMetric-Count" : "232",' +
+                '"EPA-ActivityMetric-Count": "232", "Total-ActivityMetric-Count" : "232",' +
```
mbucknell, line 68: Did the services change? I think we are only changing STORET to EPA on the presentation side. So queries won't change and neither will the expected results from the services.

server/wqp/portal_ui_blueprint/templates/cache_manager.html
```diff
@@ -51,7 +51,7 @@ <h2>Rebuild the the site cache for a data provider</h2>
         <select name="DataProvider" property="data_provider" size="1">
             <option selected="selected"> </option>
             <option value="NWIS">NWIS</option>
-            <option value="STORET">STORET</option>
+            <option value="EPA">EPA</option>
```
mbucknell, line 54: So this is a case where I believe the value won't change but we want the option name to be EPA.

server/wqp/portal_ui_blueprint/views.py
```diff
@@ -420,7 +420,7 @@ def taskstatus(task_id):
 @portal_ui.route('/manage_cache')
 @authentication_required_when_configured
 def manage_cache():
-    provider_list = ['NWIS', 'STORET', 'STEWARDS', 'BIODATA']
+    provider_list = ['NWIS', 'EPA', 'STEWARDS', 'BIODATA']
```
mbucknell, line 423: I think this needs to be restored.

server/wqp/tests/test_utils.py
```diff
@@ -16,9 +16,9 @@ def test_will_nwis_return_1(self):
         """will NWIS give back a db value of 1?"""
         assert generate_redis_db_number('NWIS') == 1
 
-    def test_will_storet_return_2(self):
-        """will STORET give back a db value of 2?"""
-        assert generate_redis_db_number('STORET') == 2
+    def test_will_epa_return_2(self):
+        """will EPA give back a db value of 2?"""
+        assert generate_redis_db_number('EPA') == 2
```
mbucknell, line 21: I think this will need to be restored too

mbucknell (changes requested):
I don't believe the services have change so we still need to use "STORET" when querying the services and this is what will be in the expected responses. I am very sure I didn't catch everything so you will want to go through this PR and make sure the changes only effect presentation.

lprivette:
Wow yes this was quite the overlook on my part. Got a little too find and replace happy 😬 

I'm pulling out the tooltip test fix, and will review/revert the need storet references on this branch (`wqp-1671`)

assets/test/js/portal_ui/views/downloadFormViewSpec.js
```diff
@@ -29,9 +29,9 @@ describe('Tests for DownloadFormView', function() {
                 '<div id="download-box-input-div"></div>' +
                 '<select id="empty-one", name="empty-one" multiple></select>' +
                 '<input id="empty-two", name="empty-two" type="hidden" />' +
-                '<select id="providers-select" name="provider" multiple><option selected value="S1">Select1</option></select>' +
+                '<input type="checkbox" name="providers" class="datasources usa-checkbox__input" value="S1" checked/>' +
```
mbucknell, line 32: At some point in your development, I would like the inputs to follow USWDS guidelines more closely. Specifically in this case there doesn't appear to be a label. If there isn't a visible label, then aria-label should be used so that screen readers can navigate the page. I don't think it needs to be dealt with write now but as you update the design of the form, we should pay attention to this.

mbucknell (approved):

assets/test/js/portal_ui/views/downloadFormViewSpec.js
lprivette: Got it.

PR #996, opened 2021-07-21
1695 toggle between adv and basic

lilyhoutman:
Before making a pull request
----------------------------

- [ ] Make sure all tests run

Title
-----------

Description
-----------
reviewing your pull request

After making a pull request
---------------------------

### DIFF
--- a/server/wqp/portal_ui_blueprint/templates/index.html
+++ b/server/wqp/portal_ui_blueprint/templates/index.html
@@ -10,9 +10,13 @@
 {% block page_content %}
 
 <!-- Basic Form -->
+<div class="tabs">
+  <div class="tab">
+    <input type="radio" id="tab-1" name="tab-group-1" checked>
+    <label for="tab-1" class="tab-label">Basic</label>
+    <div class="content">
 <div id="basicForm">
 	<div id="formIntro">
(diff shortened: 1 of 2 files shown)


server/wqp/portal_ui_blueprint/templates/index.html
```diff
@@ -270,7 +276,15 @@ <h2 class="usa-step-indicator__heading">
 		</div>
 	</div>
 </div>
+</div>
+</div>
+</div>
 
+<div class="tab">
+  <input type="radio" id="tab-2" name="tab-group-1">
```
lprivette, line 284: lets include more info in the id/name. It should indicate that it's associated with the advanced or basic tab, not just a number. Also make sure the name is correct, look like it's the same for each tab right now

lilyhoutman: i renamed some parts, but the reason that the name is the same for both is because these tabs are created through a modified version of radio buttons, which needs the same name for all of them because that groups them and allows only one to be selected at a time. naming them different things breaks the toggle. the id is what makes them unique but the name doesn't.

lprivette (approved):

PR #1013, opened 2021-08-02
1662

lilyhoutman:
Before making a pull request
----------------------------

- [ ] Make sure all tests run

Title
-----------

Description
-----------
reviewing your pull request

After making a pull request
---------------------------


assets/js/uswdsComponents/tooltip.js
```diff
@@ -0,0 +1,392 @@
...
+        : tooltipBodyOffset;
+
+    return offset;
+  };
+
+  /**
+   * Positions tooltip at the top
+   * @param {HTMLElement} e - this is the tooltip body
+   */
+  const positionTop = (e) => {
```
mbucknell, line 116: It would be better to set all of these styles in style sheets. Why is it necessary to set them here? I didn't look at the code carefully so I may be missing something. It feels a little bit like we are reimplementing the USWDS tooltip with the style changes and positioning code. Was this taken from the USWDS-Vue repo? What I have found is that they may have added tooltips prior to USWDS releasing them so they implemented themselves. I have found this to be true of the modals.

assets/js/uswdsComponents/index.js
```diff
@@ -0,0 +1,5 @@
+const tooltip = require("./tooltip");
```
mbucknell, line 1: We should not be using commonjs (require) to import modules now that we are using rollup-plugin-vue. We should be using ES6 (import/export).

assets/js/uswdsComponents/uswdsTooltip.js
```diff
@@ -2,7 +2,7 @@
  * Function adds USWDS tooltip classes to element
  * @param {Object} elem - element selection to append the tooltip
  */
-import { tooltip as uswds_tooltip } from "../../node_modules/uswds/src/js/components";
+import { tooltip as uswds_tooltip } from "./index.js";
```
mbucknell, line 5: This is a confusing pattern to put some code in tooltip and then import it into uswds_tooltip. Seems like it should be in one file.

mbucknell (changes requested):
It seems like we reimplemented the tooltip javascript rather than use USWDS. Can we justify why we implemented this way.

We also need to remove the commonjs import

lilyhoutman:
> It seems like we reimplemented the tooltip javascript rather than use USWDS. Can we justify why we implemented this way.
> 
> We also need to remove the commonjs import

So all we are doing here is copying what USWDS does for tooltips, and changing 2 words so the tooltips appear on **click** rather than **hover**. All of the styling of using two files, using require, etc was just following USWDS standard.

I agree it does seem like a lot of effort just for one change. But obviously we cannot edit the file in node modules

Do we still want the tooltips to appear on click instead of hover? If so I can work on cleaning this up

Otherwise we can delete these two added files and leave it as hover

mbucknell:
My recommendation is that we not modify USWDS components in this fashion. They become unmaintainable as they don't get updated with updates to the framework and what we want really doesn't meet the guidelines for a USWDS tooltip.

I see a few options.

- I think we may want reevaluate whether the USWDS tooltip is the right component to use for this purpose. Their recommendation is the the tooltips be brief. Many of these are not short descriptions but quite lengthy. What we have implemented before is not really a tooltip as much as popup help boxes.  USWDS doesn't have a component that does that.
- Perhaps what we can do instead is addEventListener for the click and removeEventListener for hover rather than copy the whole file. This will likely require some work with the browser debugger to see if we can access the showTooltip function.
- Use the USWDS as designed with a hover.

lilyhoutman:
> My recommendation is that we not modify USWDS components in this fashion. They become unmaintainable as they don't get updated with updates to the framework and what we want really doesn't meet the guidelines for a USWDS tooltip.
> 
> I see a few options.
> 
> * I think we may want reevaluate whether the USWDS tooltip is the right component to use for this purpose. Their recommendation is the the tooltips be brief. Many of these are not short descriptions but quite lengthy. What we have implemented before is not really a tooltip as much as popup help boxes.  USWDS doesn't have a component that does that.
> * Perhaps what we can do instead is addEventListener for the click and removeEventListener for hover rather than copy the whole file. This will likely require some work with the browser debugger to see if we can access the showTooltip function.
> * Use the USWDS as designed with a hover.

Ok, in our Wednesday meeting let's see what everyone thinks regarding these options!

My personal thoughts are I do not mind it being used as designed, with hover instead of click.

If we were to do so, I would want to remove any links from the tooltips (but honestly tooltips probably shouldn't have links anyway based on the guidelines!)

And related to the first bulletpoint, I'd also be happy to reduce the text in the tooltips even more. I did some reductions when I started on this project, but I think I'd be better equipped to do that now that I'm more familiar with the tool. We could also reach out to some of the people from our first round of user tests for recommendations on the tooltip content.

lilyhoutman:
@mbucknell I am going to keep this branch and pull request open. I'll remove the duplicated js files, and then on this branch I will edit the tooltip content, letting you know when it is complete

mbucknell (approved):

lilyhoutman:
tooltips now reduced further and links removed!

lprivette (approved):

PR #1031, opened 2021-09-17
1730 update wqp colors

lilyhoutman:
https://internal.cida.usgs.gov/jira/browse/WQP-1730

Before making a pull request
----------------------------

- [ ] Make sure all tests run

Title
-----------

Description
-----------
reviewing your pull request

After making a pull request
---------------------------


mbucknell (approved):
I would again encourage the use of USWDS colors. This makes it simpler to guarentee that we meet accessibility requirements. USWDS colors can be specified by using mixins. See https://designsystem.digital.gov/design-tokens/color/overview/

lilyhoutman:
> I would again encourage the use of USWDS colors. This makes it simpler to guarentee that we meet accessibility requirements. USWDS colors can be specified by using mixins. See https://designsystem.digital.gov/design-tokens/color/overview/

I definitely agree. I think I've been getting some conflicting tasks on my JIRA tickets (which isn't your fault). Not 100% sure where all of the ideas are coming from, but I've been told to change things like font and colors, which goes against USWDS standards. I'm going to update this branch with USWDS colors.

mbucknell:
@lilyhoutman , you can likely find USWDS colors that are close to the requested. The advantage of course is that when you need contrasting colors (like for text on the background), you can choose an appropriate USWDS color and be guaranteed that it meets WCAG guidelines.

assets/scss/custom/custom.scss
```diff
@@ -10,7 +10,7 @@ $black: #000;
 @import "../../node_modules/uswds/dist/scss/uswds";
 
 body {
-  background-color: #BDCCD4;
+  background-color: #bbc8f5; //USWDS indigo-cool-20
```
mbucknell, line 13: You should be able to use the mixin like this @include u-bg('indigo-cool-20'). Here is an example in one of my other projects where we use the mixins: https://github.com/usgs/waterdataui/blob/master/assets/src/styles/main.scss#L58

PR #1100, opened 2021-11-18
Wqp 1787 use uswds colors

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WQP - 1787 Use USWDS Colors
-----------
This converts the colors from Hex codes to use USWDS tokens and changes the main menu hover colors to white with a white underline. 

Description
-----------
Previous CSS styles frequently used the '!important' property to override some of the USWDS styling. I was able to rearrange and nest the style selectors to be more specific and eliminate the need to use !important.

I also moved some to the CSS styles from 'custom' to their own 'partials' file where it seemed to helpful to do so. 

One thing outstanding is the black chevron in title menu, it doesn't pass contrast requirements and though I have tried, currently I can't figure out how to change the color. It is only an issue when the item is not selected because the chevron turns white when the item is clicked. 

There is still a bit of rearranging of the CSS that could be done, and would be valuable, but it would require some investigation to see whether the styles require global level access or if their scope could be reduced. 

After making a pull request
---------------------------


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added search and download functionality to basic form
 - Fix download bug when switching between basic and advanced form
 - Removed jQuery, converted form to Vue and replaced Select2 with Vue Multiselects
+- Changed colors to from hex to USWDS tokens
```
mbucknell, line 19: I don't think we should update the Changelog anymore. When we make the major release of this, I will summarize the changes, but they are too numerous to list here and the current Changelog text does not capture the changes that were made.

server/wqp/portal_ui_blueprint/templates/index.html
```diff
@@ -53,7 +53,7 @@ <h3 class="usa-alert__heading">Welcome to the new Water Quality Portal</h3>
       <label for="basic-tab" class="tab-label" v-on:click="syncAtoBForm()">Basic</label>
       <div class="tab-panel">
         <div class="tab-content">
-          <div id="basicForm">
+          <div id="basic-form">
```
mbucknell, line 56: It's actually funny. I changed the id in the <form> tag to kebab case.

mbucknell (approved):
Yeah .. merging is going to be great fun!

PR #1111, opened 2021-11-22
Wqp 1782 only site profile is downloading

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

Title WQP-1782 Only Site Profile is Downloading
-----------
When using the 'basic' form, all the downloads were for 'station' regardless of the 'Data Profile' chosen.  This corrects that change.

Description
-----------
The root of the problem is that the form has two instances (basic and advanced) and only the 'advanced' instance form action attribute was getting updated when the Data Profile radio button was changed. After trying many options, I opted to simply change the form action attribute on both forms at the same time to the same form action. Since the form action is the same for both forms. 
  

After making a pull request
---------------------------

### DIFF
--- a/assets/js/DownloadProgressDialog.vue
+++ b/assets/js/DownloadProgressDialog.vue
@@ -83,6 +83,7 @@ export default {
         }
     },
     updateProgress(counts, resultType, fileFormat, continueFnc) {
+      console.log(`in update progress with resultType ${resultType}, fileFormat ${fileFormat}`)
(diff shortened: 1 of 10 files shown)


mbucknell (approved):
I think this is the best we can  do at the moment. Thanks for adjusting the css. For the most part, I got it working the same as before with a few easy tweaks. It looks like you have found some more easy tweaks to make.

PR #1117, opened 2021-11-30
Handle the keydown for the enter key when hit inside the NLDI map div…

mbucknell:
… so that if a user hits enter it is handled here and doesn't bubble up to the accordion. Also cleaned up the NLDIView component.

Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Hitting enter within the NLDI component no longer closes the "Select Location Parameters" accordion.

Description
-----------

After making a pull request
---------------------------

### DIFF
--- a/assets/js/views/NldiView.vue
+++ b/assets/js/views/NldiView.vue
@@ -1,6 +1,3 @@
-<template>
-</template>
-
 <script>
 import log from 'loglevel';
 import partial from 'lodash/partial';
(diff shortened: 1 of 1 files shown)


assets/js/views/NldiView.vue
```diff
@@ -140,186 +147,192 @@ export default {
...
 
-        this.map.addControl(searchControl);
-        this.map.addControl(featureSourceSelectControl);
-        this.map.addControl(L.control.layers(baseLayers, {
-            'NHDLPlus Flowline Network' : nhdlPlusFlowlineLayer,
-            'Hydro Reference' : hydroLayer
-        }, {position: 'topleft'}));
-        this.map.addControl(clearControl);
-        this.map.addControl(L.control.zoom());
+      // Try to prevent return from bubbling up
```
mbucknell, line 272: Apologies again for reformating and lint updating. I did fix an issue where a props was mutated, replaced with a data element. The event handler vacuums up any return key presses and does not allow them to bubble up to the accordion.

abriggs-usgs (approved):
Looks good to me.

PR #1123, opened 2021-12-02
Wqp 1793 last content in footer

abriggs-usgs:
Before making a pull request
----------------------------

- [x] Make sure all tests run

WQP-1793 Fix 'Last Content Change' in footer and related 'New User Interface' alert close issue
-----------
The footer now displays the last update of the ETLs correctly on all pages and the alert banner closes

Description
-----------
SInce the web pages use Jinja2/Python for templating there isn't any native JavaScript on them. The JavaScript functions have to be added to each page in the form of a 'bundle' of code. Adding JavaScript code to a page increases load time, so we don't want to add any additional JavaScript to pages where it will not be used. This was where the errors originated, the JavaScript needed to run the ETL update times and close the alert was not included on all the pages. 

The solution was to take out the code that ran these functions (well the close function didn't have any working code--will explain later) and add this code to a separate small JavaScript bundle that could be added to each page. 

This fix worked well for the ETL update times, however as mentioned above this didn't for for the close alert. Here are the reasons:
- the alert close had a 'onclick' JavaScript event handler on the element but no matching function in the code.
- The alert close used a class from USWDS designed for use with modals to attach the JavaScript that removed alert
- Since the code used to remove the alert used the USWDS, if that JavaScript bundle wasn't added the close wouldn't (and didn't work)
Solution 
- I didn't want to add the loading time to add the entire USWDS to every page just to close the alert so I removed the class that called the USWDS (plus, since this wasn't a modal, using a modal class didn't seem right)
- I added to the JavaScript bundle for the ETL update a small JavaScript function that now does the work of removing the modal
- I changed the styles for the close button so that (as per USWDS guidance) it is at the end of the code section (for screen readers) but the uses the CSS styles to place and style the 'close X' correctly on the page

After making a pull request
---------------------------


mbucknell (approved):
Looks great!

PR #1181, opened 2022-01-05
There are many places where the query parameter has to be specified c…

mbucknell:
…orrectly.

Before making a pull request
----------------------------

- [X] Make sure all tests run

Title
-----------
Fixed query parameter for characteristic group

After making a pull request
---------------------------


abriggs-usgs (approved):
Looks like an issue that unit tests would have caught. At least it was a minor fix.

mbucknell:
Maybe. It's not clear to me whether the "name" in the template matters but I choose not to experiment but rather try to keep things consistent.

